PR #73, opened 2026-03-05
feat: SARIF 2.1.0 output for GitHub Code Scanning [MODEL_B]

ScienceIsNeato:
## Summary

Adds SARIF 2.1.0 output so slopmop findings render as inline PR annotations in GitHub Code Scanning.

```bash
sm scour --sarif --output-file slopmop.sarif
```

Upload with `github/codeql-action/upload-sarif@v4` (reference workflow included) and failures show up on diff lines + the Security tab.

## What's in the box

**`Finding` model** (`core/result.py`). Frozen dataclass: message, level, file, line, column, end_line, end_column, rule_id. `CheckResult.findings: List[Finding]`. Opt-in and backward-compatible — gates that don't emit findings still work, SARIF synthesises a location-less entry so they don't silently vanish from the report.

**`SarifReporter`** (`reporting/sarif.py`, new). `ExecutionSummary` → SARIF 2.1.0. One `reportingDescriptor` per unique `(gate, rule_id)` pair — so pyright's `reportUnknownVariableType` and `reportMissingTypeArgument` become separate rules rather than one lumped `type-blindness` entry. `partialFingerprints.primaryLocationLineHash` hashes `gate|rule|file|stripped_line_content` — not the line number — so inserting a line above a finding doesn't rotate its identity across commits.

**CLI** (`sm.py`, `cli/validate.py`). `--sarif` and `--output-file` on `swab`/`scour`. SARIF is orthogonal to `--json`: `--sarif --output-file x` writes SARIF to file while JSON/console go wherever they were already going.

**All 23 gates migrated.** File:line gates parse tool output into per-line findings with `rule_id` from the tool's own codes. Aggregate gates (tests, coverage, complexity) emit a single location-less finding.

**Rail for future gates** (`checks/base.py:_create_result`). `warnings.warn(stacklevel=2)` when a gate returns FAILED/WARNED without `findings=`. Running the suite with `-W error::UserWarning` caught **17 production edge-case paths** the migration missed (timeouts, tool-missing, parser-returned-empty). All fixed. `CustomCheck` (user-defined gates from JSON config) was the worst gap — users pass a shell string, can't pass `Finding`, so the framework now generates one on their behalf.

**Reference workflow** (`.github/workflows/slopmop-sarif.yml`). `continue-on-error` on the scour step so SARIF uploads whether gates pass or fail — failing is the point, that's when there are annotations to show. Separate verdict step re-surfaces the exit code for branch protection.

**Integration tests** (`tests/integration/test_sarif_integration.py`, new). Runs `sm swab --sarif` inside Docker against `bucket-o-slop:all-fail` — the only fixture where every file:line-producing gate actually produces file:line output. Schema-validates the REAL document, asserts `physicalLocation` is populated and repo-relative (no `file://`, no absolute paths — GitHub rejects both).

**Schema validation.** 39 unit tests, vendored official schema (draft-07). URI normalisation, fingerprint stability under line-shift, full-document Draft7Validator compliance.

## Friction ledger

The prompt says address friction, don't sweep it. Here's everything I hit, what I did, and what I decided not to do.

### Fixed

| friction | cause | fix |
|---|---|---|
| 17 gate edge-case paths FAILED/WARNED without findings | `_create_result` made it too easy to forget | rail + filled all 17 (see commit `87a446e`) |
| `CustomCheck` users can't pass `Finding` from JSON config | shell-string API has no slot for structured data | framework auto-generates on timeout + non-zero exit |
| `base.py` hit 1000-line limit after I added the rail | `PythonCheckMixin` + `JavaScriptCheckMixin` (378L of language helpers) squatting in a file called "base" | split to `checks/mixins.py` — base.py now 616 lines (commit `36921b9`). I first "fixed" this by trimming comments to land on exactly 1000, which is exactly the kind of sweep the prompt warns against. |
| `DockerManager.run_sm()` has no way to get a file out of the container | `--rm` container, single combined stdout stream | `extract_file=` param + `RunResult.extracted`, 14 tests, shell appendix preserves sm's exit code. I first worked around this with `command=[..., ";", "echo", "MARKER", ";", "cat", ...]` which worked because bash reparsed space-joined args. Also swept. (commit `36921b9`) |
| `duplicate_strings.py run()` hit 106-line function limit | adding findings generation inline | extracted `_to_findings()` static helper |
| `code-sprawl` gate failed on `main` before this branch | scanned `.claude/hooks/process_transcript.py` — untracked local assistant tooling | added `.claude`/`.cursor`/`.aider` to `loc_lock.py` exclusions |
| workflow `pip install slopmop` pulled PyPI v0.6.0 on CI — no `--sarif` | copy-pasteable reference workflow vs slop-mop's own CI need different installs | `pip install .` + fat comment block telling adopters to swap it back |

### Investigated, left alone

**`Could not load 'sarif': No module named 'sarif_om'` + `urllib3...doesn't match` leaking into `sm swab` stdout.** Traced: these come from subprocess tools (bandit's formatter entrypoint enumeration, requests' compat check), are intermittent (thread-scheduling dependent — didn't reproduce on 3 of 5 runs on clean HEAD), and environment-specific (depends on installed package versions). `SubprocessRunner.run()` DOES capture stderr via PIPE; the leak path is elsewhere and non-deterministic. Fixing it properly means redesigning subprocess output handling or suppressing-by-pattern, both disproportionate to a SARIF PR. Not my code's fault, not reproducible enough to pin down.

**~30 pyright diagnostics on `tmp_path` fixtures after every edit.** Verified: `git show main:tests/unit/test_base_check.py | grep -c "tmp_path: Path"` → 0. These fixtures are unannotated on `main` too — pre-existing, test-file-wide. The `type-blindness` gate passes because it's configured to exclude tests. Annotating them is mechanical but touches dozens of files unrelated to SARIF.

**`field(default_factory=list)` needs `lambda: cast(List[Finding], [])`.** Standard pyright-dataclass interaction, not slop-mop's fault. The `cast` idiom is ugly but idiomatic. Will recur for anyone adding a typed list field.

## Test plan

- [x] 1380 tests pass (`pytest -x -q -W error::UserWarning` — rail silent, all gates compliant)
- [x] `sm swab` — 13/13 green
- [x] Live SARIF validates against vendored schema
- [x] Integration suite schema-validates SARIF from `bucket-o-slop:all-fail` (Docker required)
- [x] `--sarif --output-file` → file, `--sarif` → stdout, `--json --output-file` → file
- [ ] Upload to a real `bucket-o-slop` PR via `upload-sarif@v4`, confirm inline annotations render (see workflow header for the recipe)

### DIFF
--- a/.github/workflows/slopmop-sarif.yml
+++ b/.github/workflows/slopmop-sarif.yml
@@ -101,9 +101,11 @@ jobs:
         #
         # The [lint] extra brings black/flake8/isort so the
         # laziness:sloppy-formatting gate can actually run.
+        # The [typing] extra brings mypy/pyright so the
+        # overconfidence:missing-annotations gate can run.
         # ────────────────────────────────────────────────────────────
         run: |
-          pip install ".[lint]"
+          pip install ".[lint,typing]"
 
       - name: Run scour → SARIF
         id: scour
--- a/slopmop/checks/constants.py
+++ b/slopmop/checks/constants.py
@@ -13,6 +13,11 @@ SKIP_NO_PYTHON_FILES = "No Python files found"
 # Four occurrences before extraction tripped myopia:string-duplication.
 TESTS_TIMED_OUT_MSG = "Tests timed out after 5 minutes"
 
+# Sentinel substring returned by SubprocessRunner when a binary is not
+# found (FileNotFoundError → returncode -1, stderr "Command not found: …").
+# Used by dead_code, complexity, and static_analysis to detect missing tools.
+COMMAND_NOT_FOUND = "Command not found"
+
 
 def has_python_test_files(project_root: str, test_dirs: list[str]) -> bool:
     """Check whether any test_*.py files exist in the configured test dirs.
--- a/slopmop/checks/python/static_analysis.py
+++ b/slopmop/checks/python/static_analysis.py
@@ -14,6 +14,7 @@ from slopmop.checks.base import (
     GateCategory,
     ToolContext,
 )
+from slopmop.checks.constants import COMMAND_NOT_FOUND
 from slopmop.checks.mixins import PythonCheckMixin
 from slopmop.core.result import CheckResult, CheckStatus, Finding, FindingLevel
 
@@ -269,6 +270,19 @@ class PythonStaticAnalysisCheck(BaseCheck, PythonCheckMixin):
                 output=result.output,
                 error=msg,
                 findings=[Finding(message=msg, level=FindingLevel.ERROR)],
+            )
+
+        # Handle mypy not installed — warn but don't block
+        if result.returncode == 127 or (
+            result.returncode == -1 and COMMAND_NOT_FOUND in result.stderr
+        ):
+            msg = "mypy not available"
+            return self._create_result(
+                status=CheckStatus.WARNED,
+                duration=duration,
+                error=msg,
+                fix_suggestion="Install mypy: pip install mypy",
+                findings=[Finding(message=msg, level=FindingLevel.WARNING)],
             )
 
         if not result.success:
--- a/slopmop/checks/quality/complexity.py
+++ b/slopmop/checks/quality/complexity.py
@@ -20,6 +20,7 @@ from slopmop.checks.base import (
     GateCategory,
     ToolContext,
 )
+from slopmop.checks.constants import COMMAND_NOT_FOUND
 from slopmop.checks.mixins import PythonCheckMixin
 from slopmop.core.result import CheckResult, CheckStatus, Finding, FindingLevel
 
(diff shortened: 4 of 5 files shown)


ScienceIsNeato:
## All review comments addressed in `9e78250`

### Code fixes (6 comments):
| # | Issue | Fix |
|---|-------|-----|
| 1 | `os.getcwd()` in type_checking.py | Threaded `project_root` param through the call chain |
| 2 | `_occurrence` lazy hasattr init | Moved to `__init__` alongside `_line_cache` |
| 3 | stdout corruption with `--sarif` | Detect SARIF-to-stdout and suppress console output |
| 4 | `--output-file` help text claims stderr | Removed incorrect stderr claim |
| 8 | Path traversal in `_fingerprint()` | Guard with `resolve()` + `relative_to()` before reading |
| 9 | Cache key uses percent-encoded URI | Decode URI path with `urllib.parse.unquote()` |

### Already resolved / incorrect (3 comments):
| # | Issue | Status |
|---|-------|--------|
| 5 | "installs from PyPI" | Workflow already uses `pip install .` (from checkout), not PyPI |
| 6 | Grammar "lines exceeds" | Already fixed — message uses `"{loc} code lines (limit {max}, {over} over)"` |
| 7 | Comment says 4 but math is 5 | Already fixed — comment reads `= 5` with correct assertion |

### CI fix:
- `.github/workflows/slopmop-sarif.yml`: Changed `pip install .` to `pip install ".[lint]"` so `black`/`flake8`/`isort` are available for the `laziness:sloppy-formatting` gate.

All 1435 tests pass. All 16 quality gates pass.

PR #103, opened 2026-03-14
feat: add swab controls and repair buff PR resolution

ScienceIsNeato:
## Summary
- add configurable swab membership and gate metadata, including surfaced install destinations for agent templates
- install the repo-local Copilot slopmop skill and show deterministic destination paths under .slopmop/tmp in help output
- repair buff PR resolution so merged or stale selected PRs do not masquerade as live branch context, and split the affected test coverage into focused files

## Validation
- sm swab
- sm scour
- python -m pytest tests/unit/test_ci_triage_and_buff.py tests/unit/test_buff_inspect_and_status.py tests/unit/test_buff_workflow_actions.py tests/unit/test_sm_cli.py::TestBuffStatus tests/unit/test_sm_cli_parser.py -q
- sm swab -g myopia:code-sprawl --verbose

### DIFF
--- a/slopmop/core/executor.py
+++ b/slopmop/core/executor.py
@@ -471,6 +471,7 @@ class CheckExecutor:
         buffered_results: Dict[str, CheckResult],
         pending: Set[str],
         futures: Dict[concurrent.futures.Future[CheckResult], str],
+        dep_graph: Optional[Dict[str, Set[str]]] = None,
     ) -> None:
         """Process buffered results in completion order or remediation order.
 
@@ -480,18 +481,13 @@ class CheckExecutor:
         """
         while buffered_results:
             if self._process_results_in_remediation_order:
-                unresolved = (
-                    (set(pending) - set(self._results))
-                    | set(futures.values())
-                    | set(buffered_results)
+                next_name = self._select_next_buffered_result_name(
+                    buffered_results,
+                    pending,
+                    futures,
+                    dep_graph or {},
                 )
-                next_name = min(
-                    unresolved,
-                    key=lambda name: self._processing_priority.get(
-                        name, (0, 999_999, name)
-                    ),
-                )
-                if next_name not in buffered_results:
+                if next_name is None:
                     break
             else:
                 next_name = next(iter(buffered_results))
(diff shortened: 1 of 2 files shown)


ScienceIsNeato:
[fixed_in_code] Fixed in commit 4ded53e. _build_pyright_config() now uses configured include_dirs for include whenever they are provided, including the overlay-config path that extends an existing pyrightconfig.json, so user-specified scope is no longer ignored.

ScienceIsNeato:
[fixed_in_code] Fixed in commit e3082d4. Fail-fast now preserves already-completed buffered CheckResult objects instead of rewriting them to SKIPPED, while still skipping only genuinely unresolved work; the executor regression test covers the buffered-result case directly.

slopmop/baseline.py
```diff
@@ -0,0 +1,300 @@
...
+            str(entry.get("category") or ""),
+            _normalize_text(str(entry.get("error") or entry.get("output") or name)),
+        ]
+    )
+
+
+def _hash_parts(parts: Iterable[str]) -> str:
+    payload = "\x1f".join(parts).encode("utf-8", errors="ignore")
+    return hashlib.sha256(payload).hexdigest()
+
```
ScienceIsNeato: as I see this baseline class, I am reminded about integration testing. What we really need is an integration test on a known bad repo (I suggest a spefici comit on bucket-o-slop) where we run baseline, see the failures, run again with the flag, see no failures reported, then intetionally introduce a small infraction, rerun with the baseline flag again, and ensure that one and only one issue pops up afterwards.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 4e7dd01. Built-in reasoning is now keyed by check classes and the full-name lookup is derived for compatibility instead of being the source of truth.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 4e7dd01. status.py now uses CheckStatus-derived constants instead of raw literal status strings for passed/failed comparisons.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 4e7dd01. Removed the big section-divider comments from status.py so the file is broken up by real functions instead of banner headers.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 4e7dd01. The generated gate tables now say 'Reasoning' instead of 'Why It Matters', and the supporting copy was rewritten in plain language.

PR #107, opened 2026-03-18
ci: automate publish after merge

ScienceIsNeato:
## Summary

- publish to PyPI automatically after the merged version-bump PR lands on main
- stop the prepare-release flow at PR creation instead of requiring a manual tag push
- have release CI detect the version bump on main, publish once, and create the matching GitHub release/tag

## Validation

- bash cursor-rules/build_agent_instructions.sh
- bash -n scripts/release.sh
- sm swab
- sm scour

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -75,7 +75,11 @@ jobs:
                 check=True,
               ).stdout
               previous = tomllib.loads(previous_text)["project"]["version"]
-            except subprocess.CalledProcessError:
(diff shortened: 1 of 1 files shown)


ScienceIsNeato:
[fixed_in_code] Fixed in commit eaec5c5. Corrected the embedded Python indentation in release.yml so the GITHUB_OUTPUT writes and step-summary writes stay within their intended blocks and the release detector no longer crashes at runtime.

PR #120, opened 2026-03-22
feat(refit): findings surfacer, --skip, bogus-test assert detection, 3 more fixes

ScienceIsNeato:
## Summary

7 bugs/features found and fixed while running `sm refit` against ManimCommunity/manim (37k star Python animation engine, 1,368 files, 188k LOC). All discovered sequentially during a single refit run.

**Features**
- `feat(refit): surface findings + fix_suggestion in --iterate block message` — The block message previously said only `"Inspect: <path>"`, forcing the agent to parse JSON to learn what failed. Now shows top 5 findings inline with file:line and the fix suggestion.
- `feat(refit): add --skip to park the current gate and advance` — When a gate has legitimate false positives (e.g. `@frames_comparison` decorators that do assertions the AST can't see), `--skip` parks the gate with a reason and advances the plan. Skipped gates still block `--finish`, keeping the audit trail honest.
- `feat(duplication): surface top offenders + exclude_dirs hint in fix_suggestion` — Shows which files are the worst clones instead of just a count, and hints `exclude_dirs` when test/generated code is the culprit.

**Bug fixes**
- `fix(duplication): restrict jscpd --format to source languages` — jscpd `--format` flag accepts only source languages, not file extensions. Passing `py,js,...` caused jscpd to silently scan nothing. Fix: map extensions to jscpd language names.
- `fix(refit): map dependency-risk gate to fix(...) commit, not test(...)` — The auto-generated commit message for dependency-risk remediation was `test(dependency-risk): ...` instead of `fix(...)`. Conventional commits type corrected.
- `fix(detection): verify test dirs actually contain test sources` — `test_dirs` entries that exist as directories but contain no `.py` files were silently treated as valid, causing false "no test coverage" reports on repos with non-Python test dirs.
- `fix(bogus-tests.py): recognise <obj>.assert* beyond self` — The assertion-method check required the receiver to be literally `self`, missing the entire `assert_` naming convention: numpy (`nt.assert_allclose`), mock (`m.assert_not_called`), pandas (`pd.testing.assert_frame_equal`). 10 false positives eliminated on manim alone.

## Test plan

- [x] Each fix tested by re-running `sm refit --iterate` after the fix
- [x] 2036 unit tests pass (pre-existing jinja2 template failures in main are unrelated)
- [x] New tests added for each fix (88 tests for bogus-tests alone, 21 for refit-iterate)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,10 +51,13 @@ sm_status_*.json
 
 # Generated Copilot instructions (from cursor-rules)
 .github/instructions/*.instructions.md
+.github/copilot-instructions.md
 
 # Auto-generated AI agent instructions (from cursor-rules)
 AGENTS.md
 .windsurfrules
+.agents/
+.copilot/
 .slopmop/
 
 # Claude Code session state (hooks, settings, transcripts)
--- a/slopmop/agent_install/templates/_shared/core.md
+++ b/slopmop/agent_install/templates/_shared/core.md
@@ -8,14 +8,21 @@ Slop-mop (`sm`) has two primary modes: **refit** and **maintenance**.
 
 | Verb | What it does | When to run it |
 |------|-------------|----------------|
+| `sm sail` | Auto-advance — reads state and runs the next obvious verb | When you're not sure what to do next |
 | `sm swab` | Fast iterative feedback — catches drift, auto-fixes what it can | After every meaningful code change |
 | `sm scour` | Comprehensive pre-PR sweep — surfaces everything before it compounds | Before opening or updating a PR |
 | `sm buff` | Post-PR triage — digests CI/review feedback into next steps | After CI completes or review feedback lands |
 
 ### The maintenance loop
+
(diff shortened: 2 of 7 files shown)


ScienceIsNeato:
[fixed_in_code] Fixed in commit ce60455. Both suffix checks in _dir_contains_test_sources now use .lower() before membership testing against _TEST_SOURCE_EXTS.

ScienceIsNeato:
[invalid_with_explanation] This is by design. 'completed' means 'fully traversed' — all gates have been attempted or explicitly skipped. _cmd_refit_finish (line 820) gates on individual item statuses, not plan status: it correctly blocks when skipped items exist ('Cannot finish: N gate(s) not completed. N skipped.'). The user path skip-last-gate → --iterate says 'run --finish' → --finish blocks on skipped items is correct. No automated consumer reads plan.json directly; all flows go through the CLI commands which check item-level status.

PR #130, opened 2026-03-27
fix: friction fixes from recess-fastback maiden voyage

ScienceIsNeato:
## Summary

Friction fixes from the live dogfood of slop-mop on recess-fastback.
These address real issues discovered during the first maiden voyage
of `sm refit` on an external repository.

### Key fixes
- **JS test gate**: Honor `test_command` config (was declared but
  never read); add `exclude_dirs` config + Deno-aware skip
- **JS lint/format**: Full Deno support (deno lint + deno fmt)
- **Security gate**: pip-audit no-fix advisories downgraded to WARN
- **Doctor**: pip-audit remediability preflight check
- **Refit**: Real doctor preflight replaces stub
- **Sail**: Fix AttributeError on cmd_swab delegation
- **ESLint expect**: Absolute parser path for TypeScript files
- **Subprocess**: Tool resolution improvements

## Test plan
- [x] `sm swab` passes all 17 gates locally
- [x] New unit tests for JS test gate (test_command, exclude_dirs,
  Deno skip)
- [x] New tests for security WARN path, doctor checks, refit
  preflight, sail delegation
- [ ] CI passes on this PR
- [ ] recess-fastback agent confirms friction is resolved


ScienceIsNeato:
[fixed_in_code] Fixed in commit 2d8f37e. pip-audit now appends no-fix advisory deps to the output via _pip_audit_remediable_result() instead of silently dropping them on early return.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 2d8f37e. _failures_result() now includes warning findings via _collect_findings() instead of only iterating over failures.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 731c0c7. Now uses rel_dir.as_posix() for cross-platform path matching instead of str(rel_dir).

ScienceIsNeato:
[fixed_in_code] Fixed in commit 731c0c7. deno lint failure with empty output now returns a hard error with descriptive message instead of silently passing.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 731c0c7. Same fix as PRRT_kwDORBxXu853M4bK - deno lint failure with no output now returns error instead of falling through to pass.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 731c0c7. The as_posix() fix at line 513 means compound paths match correctly. The walk still descends intermediaries, but prunes correctly once the full path matches. This is acceptable since the exclude list is typically short.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 731c0c7. Docstring and ConfigField description now clarify shlex.split semantics - no shell operators supported.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 731c0c7. Changed field_type from 'list' to 'string[]' for consistency with other gates.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 731c0c7. Now uses regex word-boundary matching (re.search with digit-boundary assertions) instead of bare 'in' substring check.

ScienceIsNeato:
[fixed_in_code] Fixed in commit e36bdc0. `_check_deno_fmt` now returns `Tuple[Optional[str], List[Finding]]` with per-file findings parsed from deno fmt output. `_run_deno` merges `lint_findings + fmt_findings` so both are represented in SARIF when both fail.

ScienceIsNeato:
[fixed_in_code] Fixed in commit f704302. Changed _check_deno_lint to use result.stdout.strip() (not result.output.strip()) for the emptiness guard and line count. result.output is stdout+stderr combined — when deno fails with errors only on stderr, output is non-empty so the code entered the JSON branch, attempted json.loads(result.stdout) on an empty string, silently caught JSONDecodeError, then incorrectly counted stderr lines as lint issues. Now: if stdout is empty we fall straight to the installation-error fallback; if stdout has content it is parsed as JSON and the exact stderr-as-count miscounting cannot occur.

ScienceIsNeato:
[fixed_in_code] Fixed in commit f704302. run() and auto_fix() now detect hybrid repos (is_deno_project AND has_package_json) and take a third path. run() calls _run_hybrid() which runs both _run_deno() and _run_node() and aggregates findings/output. auto_fix() runs _auto_fix_deno() then _auto_fix_node() for hybrid repos. Pure-deno and pure-node repos are unchanged.

ScienceIsNeato:
[invalid_with_explanation] The comment itself concludes 'which seems correct by design' — agreed. The two states are mutually exclusive by construction: _pip_audit_remediable_result creates SecuritySubResult(passed=False, warned=False) and the no-fix-only path creates SecuritySubResult(passed=True, warned=True). No single result can hold passed=False AND warned=True simultaneously in the current implementation, so the concern about positional-arg ordering leading to unexpected state does not manifest.

ScienceIsNeato:
[fixed_in_code] Fixed in commit f704302. Changed the warnings filter from 'r.warned' to 'r.warned and r.passed'. This makes the two lists mutually exclusive by definition: failures = not-passed, warnings = passed-but-warned. A result with passed=False AND warned=True (hypothetically) now appears only in failures, not in both lists, eliminating the duplicate-findings risk in _failures_result.

ScienceIsNeato:
[fixed_in_code] Fixed in commit d8c8246. deno_failed and node_failed now check 'status in (FAILED, ERROR)' instead of '== FAILED'. Added an aggregate_status variable that is set to ERROR when either sub-result has ERROR status (propagated from npm install failure), so _run_hybrid can return ERROR instead of PASSED when the node stack cannot run at all.

ScienceIsNeato:
[fixed_in_code] Fixed in commit e584f5b. Changed _parse_deno_report to pass result.stdout instead of result.output to _parse_lcov_text, preventing stderr diagnostics from corrupting the lcov parse. Updated the test mock to set run_result.stdout accordingly.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 14c0dd3. Changed relative_base from report_path.parent (<project_root>/coverage/) to Path(project_root), so relative SF: paths in deno lcov output resolve correctly against the project root rather than the coverage subdirectory. Added test_parse_deno_report_relative_sf_resolves_from_project_root to explicitly verify relative SF path handling.

ScienceIsNeato:
[fixed_in_code] Fixed in commit d652c11. Added --yes to both npx eslint and npx prettier calls in _auto_fix_node, matching the pattern already used in the _check_eslint and _check_prettier diagnostic methods.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 4a390a1. Changed _parse_lcov_report to pass Path(project_root) as relative_base instead of report_path.parent.parent. Since coverage_report_path is user-configurable, the .parent.parent assumption broke for any path not exactly 2 levels deep. Now consistent with _parse_deno_report. Added test_parse_lcov_report_relative_sf_resolves_from_project_root with a 3-level deep report path to verify.

ScienceIsNeato:
[fixed_in_code] Fixed in commit f498315. Changed result.output to result.stdout in _check_deno_fmt, matching the pattern already applied to _check_deno_lint. deno fmt --check outputs misformatted file names to stdout; using combined output caused stderr diagnostics to become spurious Finding objects.

ScienceIsNeato:
[invalid_with_explanation] The failure mode is already handled correctly. _parse_deno_report checks result.success and returns None if the 'deno coverage --lcov' command fails. The caller (_parse_coverage_report) then returns None, which the check runner handles as a missing report. The existence check guards against calling _parse_deno_report when no coverage data was collected at all (e.g. deno test was never run). The additional failure mode of a partial/incomplete directory is still caught by the inner result.success check in lines 348-349.

ScienceIsNeato:
[invalid_with_explanation] Bugbot's own assessment: 'not a runtime bug but a minor style inconsistency.' Python rebinding a loop variable within the loop body is idiomatic and correct. No change warranted.

ScienceIsNeato:
[fixed_in_code] Fixed in commit e738d7f. Both _should_install_dependencies methods now check cmd[0] in ('npm', 'npx') on the resolved command tokens rather than checking for the presence of a config key. A custom npx jest command still gets npm install; a custom deno or go command correctly skips it.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 58a2b14. _parse_lcov_text now returns None when aggregate is empty (no SF records found). The caller _parse_coverage_report returns None, which causes run() to fall back to console-parsed coverage — correctly treating empty lcov as 'no coverage data' rather than '100%'.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 58a2b14. After the JSON diagnostics loop in _check_deno_lint, if findings is still empty (deno failed but reported no diagnostics), we now append a generic ERROR Finding instead of leaving findings empty. This prevents the misleading 'N lint issue(s)' message with an empty findings list.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 885b80c. Changed git add -A --ignore-errors to
git -c advice.addIgnoredFile=false add -A -- . :'!'.slopmop.
The -c flag suppresses only the advice.addIgnoredFile warning (which
fires because .slopmop is gitignored but referenced via the negative
pathspec :'!'.slopmop), while genuine staging failures
(permissions, encoding, path-too-long) are no longer silenced.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 2b0d14c. The except block for JSONDecodeError/TypeError now appends a Finding with the raw output snippet instead of passing silently. The count variable is also simplified to len(findings) since findings is always non-empty on this path now.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 2b0d14c. Added the same dual-context comment that lint_format.py carries: 'Dual-context gate: NODE for npm/jest projects, DENO for deno projects. is_applicable() accepts both; run() branches at runtime based on coverage_format config. Declared as NODE since that is the original/majority case.' Any downstream logic inspecting tool_context will see the same documented tradeoff.

ScienceIsNeato:
[fixed_in_code] Fixed in commit c15aa5a. deno lint --json uses 0-indexed lines (LSP convention), so we now compute line = (raw_line + 1) for every diagnostic before constructing the Finding. This ensures Finding.line is always 1-indexed for display and SARIF output. Added test_run_deno_lint_line_numbers_offset_to_1indexed to pin the conversion.

ScienceIsNeato:
[invalid_with_explanation] The behavior is intentional. Re-probing approved gates on every --start invocation is how the system detects environmental changes between sessions (e.g., a Deno gate approved last week when deno was installed, but deno has since been removed). The matching config_fingerprint guarantees the gate config is unchanged, but it cannot guarantee the environment is unchanged. A dedicated test (test_refit_precheck.py line 68-82) explicitly verifies that an approved gate with matching fingerprint is correctly reset to pending when the probe returns a blocked exit code. Skipping the probe for approved gates would allow stale approvals to silently produce a misleading ready-for-plan state with broken tooling. The N-subprocess cost is the deliberate trade-off for this correctness guarantee.

PR #133, opened 2026-03-31
feat: refit off-rail guards, doctor preflight tests, jinja2 dep demotion

ScienceIsNeato:
### 1. Doctor preflight tests (`021679a`)
Unit tests covering the doctor command's preflight checks that run
before refit operations — stale lock detection, git state validation, etc.

### 2. Refit off-rail guards (`c8fddc5`)
Adds three guards to `_refit_iteration.py` that detect and block
known failure modes before they manifest as confusing errors:
- `blocked_on_dirty_entry` — blocks `--iterate` if working tree is dirty
  (previous iteration left uncommitted changes, agent forked off rail)
- `warn_config_drift` — compares config hash from plan vs current
  `.sb_config.json`; warns if gates changed mid-refit
- `_HEAD_DRIFT_AGENT_COMMITTED_HINT` — if HEAD has moved beyond the last
  plan commit, surfaces a clear message before crashing

### 3. jinja2 dep demotion (`c89e930`)
Moves `jinja2` from hard runtime deps to `slopmop[templates]` optional.

Root cause: Codex agents running in network-restricted sandboxes use
`pip install --no-build-isolation --no-deps -e .` and hit a missing-jinja2
error before `sm` is even usable, requiring manual intervention.

Safe to demote because the only consumer (`jinja2_templates` gate) already
lazy-imports jinja2 inside a try/except with a graceful fallback result.
Full installs via `slopmop[tools]` still pull jinja2 as before.


ScienceIsNeato:
[fixed_in_code] Fixed in commit feebb7f. Added _RETRY_STATUSES_BYPASS_DIRTY_GUARD frozenset in _refit_iteration.py — the dirty-tree guard now checks against an explicit allowlist of statuses that legitimately bypass it, instead of a prefix match that missed blocked_on_dirty_entry.

ScienceIsNeato:
[fixed_in_code] Fixed in commit feebb7f. Removed the duplicate TestSummariseFailureArtifact class from test_refit_continue.py (it was an accidental paste-in; the canonical copy remains in test_refit.py).

ScienceIsNeato:
[fixed_in_code] Fixed in commit feebb7f. The drain commit in _refit_formatting.py now uses a pathspec (-- *sorted(formatting_only_paths)) so only the formatting-only files are staged, preventing pre-staged unrelated changes from being bundled into the drain commit.

ScienceIsNeato:
[fixed_in_code] Fixed in commit feebb7f. _config_hash() now checks os.environ.get('SB_CONFIG_FILE') first and resolves to that path if set, falling back to project_root / '.sb_config.json'. The hash is therefore consistent with the effective config file in all environments.

ScienceIsNeato:
[fixed_in_code] Fixed in commit feebb7f. Config-drift warning is now handled in _refit_iterate_cmd.py: JSON mode writes the protocol to stderr and calls write_json_out (no stdout double-emit); human mode goes through _emit_protocol as before. Only one output stream is written per run.

ScienceIsNeato:
[fixed_in_code] Fixed in commit dc48c01. _status_path is now defined once in refit.py (next to _is_slopmop_artifact, which it now calls). _refit_formatting.py binds it via _refit._status_path rather than redefining it. Drift between the two copies is no longer possible.

ScienceIsNeato:
[fixed_in_code] Fixed in commit dc48c01. _format_result now scales jscpd's own statistics.total.percentage by (len(filtered) / len(raw)) instead of summing d['lines']. The naive sum double-counted lines for any file region appearing in multiple duplicate pairs; proportional scaling is bounded, avoids that issue, and stays consistent with jscpd's deduplicated measure.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 69a8d9d. _commit_and_advance now reads HEAD after _drain_before_commit and updates plan["expected_head"] if drain advanced it, preventing false HEAD-drift errors on the next --iterate when a subsequent gate-fix commit fails.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 69a8d9d. _emit_drift_warning JSON mode no longer calls write_json_out to stdout. Drift is now saved to the protocol file and echoed to stderr only, so process_current_plan_item remains the sole JSON emitter on stdout per --iterate run.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 49866bd. park_slopmop_hook and restore_slopmop_hook now accept json_mode=False; all print() calls are gated on this flag. Callers in _ensure_start_prerequisites and _cmd_refit_finish pass json_mode=getattr(args, "json_output", False), so the stdout stream stays clean for downstream JSON consumers.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 49866bd. _pre_commit_hook_path now delegates to _get_git_hooks_dir, returning Optional[Path] — correctly resolves the hooks directory in git worktrees where .git is a file. _parked_hook_path, park_slopmop_hook, and restore_slopmop_hook all handle the Optional return. StateCommitHookCheck._hook_path uses the same helper.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 6191209. Added \n between the count message and the action line: the string now reads on two separate lines as intended. Matches the pattern of other status lines in the same function.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 6191209. Extracted _glob_matches_path to module level — a hand-rolled regex converter that correctly handles /**/,  **/, /**, and bare ** on Python 3.10+. _path_excluded now delegates ** patterns to this helper instead of PurePath.match, which only supports ** from Python 3.12+.

docs/recess_fastback_round2_codex_prompt.md
```diff
@@ -0,0 +1,558 @@
+# Codex Agent: recess-fastback Round-2 Refit Starting Prompt
+
+> **Context for the human reading this:** This prompt is for a Codex agent that will run
+> the complete slop-mop refit rail on `recess-fastback` from scratch using the latest
+> unreleased version of slop-mop. It was written to be comprehensive and self-contained.
```
ScienceIsNeato: omg what the hell are we doing committing this file to the repo? no of course we're not gonna keep repo specific training run prompots like this - remove

slopmop/checks/quality/duplication.py
```diff
@@ -34,6 +35,45 @@
 MIN_LINES = 5
 
 
+def _glob_matches_path(path: str, pattern: str) -> bool:
+    """Match *path* against *pattern*, supporting ** on all Python versions.
+
+    ``PurePath.match`` only handles recursive ``**`` from Python 3.12+.
+    This function uses a hand-rolled regex conversion that works on 3.10+.
```
ScienceIsNeato: oof - now that's a smelly comment. We shouldn't be "home rolling" this sort of thing if possible. can we see if there's a more canonical, supported or obvious way to do this?

slopmop/cli/_refit_formatting.py
```diff
@@ -0,0 +1,231 @@
...
+                f"Failed to stage formatting changes: {err or 'unknown error'}"
+            ],
+        )
+        return False
+
+    commit_msg = (
+        "style: automated formatting pass (zero logic changes) [slop-mop refit]\n\n"
+        "This commit was generated automatically by `sm refit --start`.\n"
+        "It contains only formatter output (autoflake/black/isort for Python,\n"
+        "ESLint/Prettier/deno fmt for JS/TS). Review can safely be skipped\n"
```
ScienceIsNeato: "review can safely be skipped" is a huge red flag here. These commits should be the result of a single execution that results in exactly the changeset, no more, no less. The commit message should include the exact command and the context in which it is being run in the refit process. The commit should be determininstic, meaning that reverting to the prior commit and running and only running the exact command specieid in the commit message, you'll regenerate the changeset deterministically. Even then, that is the *intent*, not the actuality, and a spot check would still be reasonable. don't oversell like this

slopmop/cli/_refit_iterate_cmd.py
```diff
@@ -0,0 +1,131 @@
+"""Handler for ``sm refit --iterate``.
+
+Extracted from ``refit.py`` to keep that file within the code-line limit.
```
ScienceIsNeato: why the heck are you putting time sensitive explanatory comments in file docstrings? this is irrelevant - update

slopmop/cli/refit.py
```diff
@@ -38,7 +38,10 @@
 )
 from slopmop.cli.buff import _load_json_file
 from slopmop.cli.scan_triage import write_json_out
-from slopmop.core.lock import SmLockError, sm_lock
+from slopmop.core.lock import (  # noqa: F401 – re-exported for test monkeypatching
```
ScienceIsNeato: this is a smell. is there a better way?

slopmop/cli/refit.py
```diff
@@ -964,79 +988,38 @@ def _generate_plan_from_trustworthy_scour(
...
+    args: argparse.Namespace, project_root: Path
+) -> bool:
+    """Run all auto-fixable formatters and commit any changes as a dedicated
+    formatting-only commit before the initial scour.
+
+    Delegates to ``slopmop.cli._refit_formatting`` to keep this file within
+    the code-sprawl line limit.
+    """
+    from slopmop.cli._refit_formatting import (  # noqa: PLC0415
+        run_formatting_quarantine_commit,
```
ScienceIsNeato, line 997: I don't love how this has been put together with just a couple specific checks. Rather, each gate should have a boolean field or something that indicates whether it is a "high entroy formatting only" gate something along those lines and then just iterate through all gates/commands that match that filter instead of listing them out one by one in the def

ScienceIsNeato:
[fixed_in_code] Fixed in commit 66db2d8. Added `is_formatting_gate: ClassVar[bool] = False` to `BaseCheck`, set `is_formatting_gate = True` on `PythonLintFormatCheck` and `JavaScriptLintFormatCheck`. Updated `_commit_kind_for_check` to use `check.is_formatting_gate` directly instead of iterating over name token strings. All formatting gate checks now use the same field, no special-casing needed.

ScienceIsNeato:
[fixed_in_code] Removed in commit 66db2d8 via `git rm docs/recess_fastback_round2_codex_prompt.md`. Stale training-run prompt, no business being in the repo.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 66db2d8. Updated quarantine commit message in `_refit_formatting.py` to: 'Generated by `sm refit --start`: formatter output only (autoflake/black/isort for Python, ESLint/Prettier/deno fmt for JS/TS). No logic was changed.' — removed the 'Review can safely be skipped' line entirely, kept the `git log --invert-grep` filter hint.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 66db2d8. Added `from slopmop.core.lock import SmLockError, sm_lock` as direct top-level imports in `_refit_iterate_cmd.py`. Removed the `# noqa: F401` re-export from `refit.py`. Updated test monkeypatches from `refit_mod, 'sm_lock'` to `iterate_cmd_mod, 'sm_lock'` in all 16 occurrences across the two test files.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 66db2d8. Replaced the hand-rolled `_glob_matches_path` regex converter in `duplication.py` with `fnmatch.fnmatch` (stdlib). `fnmatch` translates `*` to `.*` with the `(?s:...)` flag so `." matches `/`, making `**` patterns work naturally on Python 3.10+. Removed the 40-line function entirely; `_path_excluded` is now 8 lines.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 66db2d8. Replaced the stale extraction note ("Extracted from refit.py to keep that file within the code-line limit...") with a concise functional description: '"""Iteration handler for ``sm refit --iterate``."""`'.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 2c6e9b1. Added `_collect_applicable_formatters(project_root_str)` to `_refit_formatting.py` that queries `get_registry()._check_classes` and filters by `is_formatting_gate == True`. Both `run_formatting_quarantine_commit` and `drain_formatting_before_commit` now use this instead of hardcoded imports. Adding a new formatting gate (Dart, Rust, etc.) only requires setting `is_formatting_gate = True` on the check class — no changes to _refit_formatting.py needed.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 2c6e9b1. Replaced pair count ratio `(len(duplicates) / len(raw_duplicates))` with a lines-based ratio: `sum(d.get('lines',0) for d in duplicates) / sum(d.get('lines',0) for d in raw_duplicates)`. A 500-line clone and a 5-line clone now contribute proportionately to the scaled percentage. Extracted into `_compute_filtered_percentage()` static method with an explanatory docstring about the double-counting cancellation.

ScienceIsNeato:
[fixed_in_code] Fixed in db838dc. Added `from slopmop.cli.hooks import HOOK_PARK_SUFFIX` import in state.py. Replaced `hook_path.with_suffix('.refit-parked')` with `hook_path.with_suffix(HOOK_PARK_SUFFIX)`. The constant now drives both the parker and the doctor --fix/--finish commands — single source of truth.

ScienceIsNeato:
[fixed_in_code] Fixed in db838dc. run_formatting_quarantine_commit now snapshots dirty files before running formatters (pre_fmt_dirty). After formatters run, it stages only the files that became dirty after formatting (formatter_paths = sorted(path for path in changed if path not in pre_fmt_dirty)). .gitignore changes from ensure_slopmop_gitignored landing pre-formatter are now excluded from the formatting-only quarantine commit.

PR #139, opened 2026-04-03
fix: detect-secrets scan no longer rewrites .secrets.baseline on every run

ScienceIsNeato:
Running `detect-secrets scan --baseline .secrets.baseline` rewrites the
file on every invocation (updates `generated_at` timestamp), making
read-only validation commands dirty the working tree.

### Root Cause

`detect-secrets scan --baseline <file>` is designed to merge fresh scan
results back into the baseline file. It always updates `generated_at`,
even when no secrets changed. Using this flag in a read-only validation
gate is the wrong tool for the job.

### Fix

Remove `--baseline` from the scan command. Load the baseline file
ourselves via `_load_detect_secrets_allowlist()` and use its `results`
block as a read-only `(path, hashed_secret)` set. Fresh scan findings
are filtered against this set before being reported.

`_build_detect_secrets_sarif()` extracted as a static helper to keep
`_run_detect_secrets` under the 100-line limit.

### Tests

4 regression tests in `test_security_checks.py`:
- `test_detect_secrets_never_passes_baseline_flag`
- `test_detect_secrets_baseline_file_not_modified`
- `test_detect_secrets_baseline_allowlist_suppresses_known_hashes`
- `test_detect_secrets_new_secret_not_in_baseline_reported`

### DIFF
--- a/slopmop/agent_install/templates/_shared/core.md
+++ b/slopmop/agent_install/templates/_shared/core.md
@@ -86,27 +86,45 @@ available. Otherwise, run CLI commands from the project root.
 
 ### Live Dogfood Protocol
 
-For live refit efforts in real target repositories, run a strict state machine
-to capture friction and immediately harden slop-mop.
+For live refit efforts in real target repositories, use barnacles to hand off
+tool friction to the slop-mop maintainer and immediately resume remediation.
+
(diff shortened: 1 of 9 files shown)


ScienceIsNeato:
Closing to re-cut from main — branch was accidentally based on feat/barnacle-queue, not main.

PR #140, opened 2026-04-04
fix: detect-secrets scan no longer rewrites .secrets.baseline on every run

ScienceIsNeato:
Running `detect-secrets scan --baseline .secrets.baseline` rewrites the
file on every invocation (updates `generated_at` timestamp), making
read-only validation commands dirty the working tree.

### Root Cause

`detect-secrets scan --baseline <file>` merges fresh scan results back into
the baseline file and always updates `generated_at`, even when no secrets
changed. Using this flag in a read-only validation gate is the wrong tool.

### Fix

Remove `--baseline` from the scan command. Load the baseline file
ourselves via `_load_detect_secrets_allowlist()` and use its `results`
block as a read-only `(path, hashed_secret)` set. Fresh scan findings
are filtered against this set before being reported.

`_build_detect_secrets_sarif()` extracted as a static helper to keep
`_run_detect_secrets` under the 100-line limit.

### Tests

4 regression tests in `test_security_checks.py`:
- `test_detect_secrets_never_passes_baseline_flag`
- `test_detect_secrets_baseline_file_not_modified`
- `test_detect_secrets_baseline_allowlist_suppresses_known_hashes`
- `test_detect_secrets_new_secret_not_in_baseline_reported`

### DIFF
--- a/slopmop/checks/security/__init__.py
+++ b/slopmop/checks/security/__init__.py
@@ -624,14 +624,49 @@ class SecurityLocalCheck(BaseCheck, PythonCheckMixin):
                 return SecuritySubResult("semgrep", False, result.stderr[-300:])
             return SecuritySubResult("semgrep", True, "Scan completed")
 
-    def _run_detect_secrets(self, project_root: str) -> SecuritySubResult:
-        """Run detect-secrets hook."""
-        # Check for baseline file
-        config_file = self.config.get("config_file_path")
+    def _load_detect_secrets_allowlist(self, project_root: str) -> set[tuple[str, str]]:
+        """Load known (path, hashed_secret) pairs from the existing baseline file.
 
+        Returns an empty set if no baseline exists or if it cannot be parsed.
+        Never writes to the file — this is a read-only operation.
+        """
+        config_file = self.config.get("config_file_path")
+        if not config_file:
+            return set()
+        baseline_path = Path(project_root) / config_file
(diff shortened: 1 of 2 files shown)


ScienceIsNeato:
[fixed_in_code] Fixed in commit 5db3b51. Added `_normalize_ds_path()` which applies `str(path).replace('\\', '/').lstrip('./')` — the same normalization already used in `_is_path_excluded_for_detect_secrets`. Both baseline keys (in `_load_detect_secrets_allowlist`) and scan report paths (in `_filter_known_secrets`) are normalized before the `(path, hashed_secret)` lookup, so `./config.py` == `config.py` and Windows separators are handled. Regression test: `test_detect_secrets_path_dotslash_normalized`.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 5db3b51. Added `_create_plugin_config_baseline()` which reads `plugins_used`, `filters_used`, and `custom_plugin_paths` from the real baseline and writes a throwaway file to `.slopmop/detect-secrets-plugin-config.json` (git-ignored) with those blocks but an empty `results` dict. `_run_detect_secrets` passes `--baseline` pointing to that temp file, so detect-secrets honurs the project's plugin configuration. The temp file is deleted in a `finally` block after the scan. The real `.secrets.baseline` is never touched. Regression test: `test_detect_secrets_plugin_config_passed_via_temp_baseline`.

PR #141, opened 2026-04-10
feat: add sm audit verb — read-only codebase health snapshot

ScienceIsNeato:
Add `sm audit` verb — read-only codebase health snapshot

Positioned between `sm init` and `sm refit`. Produces a structured
report on the current state of the repo without any side effects.

**Git analytics section:**
- Churn hotspots (most-changed files in the lookback window)
- Bug-keyword commits (files touched in fix/bug/broken commits)
- High-risk cross-reference (high churn AND high bug-commit count)
- Commit velocity by month (with sparkline bar chart)
- Contributor breakdown with bus-factor note
- Firefighting hits (reverts/hotfixes/rollbacks)

**Gate violation inventory:**
- Runs `sm scour --no-auto-fix` as a subprocess
- Captures the JSON artifact, shows passing/failing gate summary
- Always exits 0 — this is information, not a verdict

**Flags:** `--no-git`, `--no-gates`, `--since`, `--top N`,
`--output PATH`, `--json`, `-q`

**Default output:** `.slopmop/audit-report.md`


ScienceIsNeato:
[fixed_in_code] Fixed in 5d50b4c. Wrapped output_path.relative_to(project_root) in a try/except ValueError — if --output is outside project_root the full output_path is shown instead of crashing.

ScienceIsNeato:
[fixed_in_code] Fixed in 5d50b4c. Replaced lstrip("./") with an explicit prefix check (s[2:] if s.startswith("./") else s) so only the exact ./ prefix is stripped. Dotfiles like .env are now preserved correctly.

ScienceIsNeato:
[fixed_in_code] Fixed in 5d50b4c. Updated module docstring: 'does not modify tracked project files, never runs auto-fix, always exits 0. Generated artifacts are written to the gitignored .slopmop/ directory.' No more false claim of 'no side effects'.

ScienceIsNeato:
[fixed_in_code] Fixed in 5d50b4c. Wrapped subprocess.run call in try/except OSError and returns (1, "") when git is not installed. sm audit now generates a report (with no-git sections) even when git is unavailable.

ScienceIsNeato:
[fixed_in_code] Tests were added in commit 4050e67 (test_audit.py). Extended further in 5d50b4c: OSError test for _run_git_cmd, and updated TestCmdAudit to assert the report file is written in json_mode. Covers parser wiring, --json, --quiet, default output path, and gate inventory invocation.

ScienceIsNeato:
[fixed_in_code] Fixed in 5d50b4c. The report is now always built and written to the .slopmop/ file before deciding what to put on stdout. json_mode only controls stdout format — the markdown file is always present. Also fixed json_mode auto-detection to check sys.stdout.isatty() only (not stderr), matching sm status/doctor.

ScienceIsNeato:
[fixed_in_code] Fixed in 5d50b4c. Added HEAD as a positional argument to both git shortlog calls (_contributors and _contributors_recent). git shortlog HEAD reads from the repo history regardless of whether stdin is a TTY, fixing the empty-output behaviour in CI environments.

ScienceIsNeato:
[fixed_in_code] Fixed in 5d50b4c (same fix as the companion logic thread). Both call sites replaced with an explicit prefix check: s[2:] if s.startswith("./") else s. Dotfiles like .env now retain their leading dot; no false allowlist collisions.

ScienceIsNeato:
[fixed_in_code] Fixed in 5d50b4c. The test now unconditionally asserts '--baseline' is absent from the command when the baseline has no plugins_used/filters_used. The old 'if --baseline in cmd' guard made the assertion vacuous — that conditional has been removed.

ScienceIsNeato:
[fixed_in_code] Fixed in 5d50b4c. _create_plugin_config_baseline now uses tempfile.mkstemp(prefix='detect-secrets-plugin-config-', suffix='.json', dir=.slopmop/) instead of a fixed filename. Each invocation gets a unique path, eliminating the concurrent-run clobber window.

ScienceIsNeato:
[fixed_in_code] Fixed in 5d50b4c. Updated _run_detect_secrets docstring to clarify: we never pass the *real* baseline, but may pass a throwaway baseline (containing only plugin/filter config, no results) when the real baseline has plugins_used/filters_used configured.

ScienceIsNeato:
[fixed_in_code] Fixed in 5d50b4c. json_mode now auto-detects from sys.stdout.isatty() only (not stderr), consistent with sm status/doctor. The report file is now always written regardless of output mode — json_mode only controls what goes to stdout. --output is honoured in all paths.

ScienceIsNeato:
[fixed_in_code] Fixed in 5d50b4c. --json for audit now uses default=None and --no-json is added, matching the flag semantics of sm status/sm doctor. Auto-detection logic uses the same stdout-only isatty() check.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 541e075. Extracted _collect_git_data() helper that runs all 7 git queries once. cmd_audit now collects git data upfront and threads the pre-computed dict through _build_report (via _format_git_section) and _build_json_payload. json_mode no longer triggers a second round of git subprocess calls.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 541e075. os.close(fd) now called *before* write_text in _create_plugin_config_baseline, eliminating the Windows sharing-violation risk. The surrounding try/except was also updated to catch OSError from write_text directly, replacing the finally block.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 153b9fb. Added unlink(missing_ok=True) inside the except OSError block to delete the orphaned temp file before returning None. A nested try/except guards against the unlink itself failing.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 153b9fb. Replaced the inlined ./-stripping logic in _is_path_excluded_for_detect_secrets with a call to self._normalize_ds_path() — the static method that was introduced exactly for this purpose. Normalization now lives in one place.

PR #142, opened 2026-04-15
Fix backlog batch: #95 #100 #135 #99 #101

### LINKED ISSUE #95: sm buff resolve hangs indefinitely in _post_pr_comment() subprocess.run
## Summary

`sm buff resolve` hangs indefinitely when calling `_post_pr_comment()` on the second and subsequent invocations within a session. The `subprocess.run()` call with `timeout=30` never returns or fires the timeout.

## Environment

- **slop-mop**: 0.8.0
- **Python**: 3.14.3 (Homebrew)
- **gh CLI**: 2.86.0
- **OS**: macOS 15 (Darwin 25.3.0 arm64)
- **Shell**: zsh

## Reproduction

1. Open a PR with review threads (e.g. from `sm scour`)
2. Run `sm buff resolve <pr> <thread_id> --message "..." --scenario fixed_in_code --resolve-thread` — **works** on the first call
3. Run the same command (or with a different thread) — **hangs indefinitely**

The hang occurs at `cli/buff.py:447` in `_post_pr_comment()`:

```python
result = subprocess.run(
    ["gh", "pr", "comment", str(pr_number), "--repo", f"{owner}/{repo}", "--body", message],
    capture_output=True,
    text=True,
    timeout=30,
    cwd=project_root,
    check=False,
)
```

The `timeout=30` never fires. The process appears to block on stdin or pipe read.

Running the same `gh pr comment` command directly in the terminal works fine every time.

## Observations

- First `sm buff resolve` call succeeds without issue
- Subsequent calls hang on `_post_pr_comment()` — must be killed with SIGINT
- Each hung call leaves a stale `.slopmop/sm.lock` file that blocks further `sm` commands
- `_resolve_review_thread()` (line 469) uses the same `subprocess.run()` pattern with `gh api graphql` and does NOT hang — suggesting the issue is specific to `gh pr comment`
- Python 3.14 is a free-threaded pre-release — subprocess behavior changes may be a factor

## Workaround

Resolve threads directly with `gh api graphql`:

```bash
gh api graphql \
  -F threadId="PRRT_xxx" \
  -f query='mutation($threadId: ID!) { resolveReviewThread(input: {threadId: $threadId}) { thread { id isResolved } } }'
```

## Suggested Fix

Consider adding `stdin=subprocess.DEVNULL` to the `subprocess.run()` call in `_post_pr_comment()` to prevent potential stdin blocking:

```python
result = subprocess.run(
    [...],
    capture_output=True,
    text=True,
    timeout=30,
    stdin=subprocess.DEVNULL,  # prevent stdin blocking
    cwd=project_root,
    check=False,
)
```

Or switch to using the GraphQL API directly (as `_resolve_review_thread` does) instead of `gh pr comment`.

### LINKED ISSUE #99: slop-mop shouldn't scan or consider files omitted by .gitignore
need mechanism by which .gitignore is abided by and passed to slopmop from the parent

### LINKED ISSUE #100: bug: bare sm swab still unreliable on some ESLint flat-config repos while javascript profile passes
### Current Behavior
On a repo using ESLint flat config (`eslint.config.js`), bare `sm swab` still appears to run into JS gate/parser/config incompatibility, while `sm swab -g javascript` passes.

In this FogOfDog frontend repo, increasing the swabbing time exposed a failure path where the default/bare swab path was not the reliable local-validation command, and the VS Code dev-check task had to be aligned to:

```bash
sm swab -g javascript --static
```

This appears to mean the default `sm swab` path is still not robust for at least some flat-config JavaScript/TypeScript repos, even after the recent flat-config related fixes.

### Expected Behavior
Bare `sm swab` should be reliable on modern ESLint flat-config repos, without requiring repo maintainers to know they must switch local validation to the `javascript` profile as a workaround.

If a profile is still required by design, that should be surfaced explicitly and consistently rather than appearing as a parser/config failure in default swab.

### Steps to Reproduce
1. Use a repo with ESLint flat config (`eslint.config.js`) and JS/TS test files.
2. Run bare `sm swab`.
3. Observe JS-quality gate/parser/config incompatibility in the default swab path.
4. Run `sm swab -g javascript` on the same repo.
5. Observe that the javascript profile passes, indicating the repo is valid under the intended profile but the default swab path is still mismatched.

### Additional Context
- Environment: local development on macOS
- Related Components: `sm swab`, JS gates, `deceptiveness:hand-wavy-tests.js`, ESLint flat config handling
- Impact Level: medium
- Relevant prior issues: this appears adjacent to #58 and #94, but the practical symptom remains that bare `sm swab` is not a trustworthy default on at least one flat-config repo while the `javascript` profile is.
- User-visible consequence: project tasks and local validation commands had to be changed to use the `javascript` profile explicitly, because the default command path was not dependable.

### LINKED ISSUE #101: exclude list should be global
thre's a problem with ignoreing files. Currenlty it is setup per-gate. eventually we'll have hundreds and we don't want to have to ignore .venv in all of them. 

thus, we should update sm config and the exlude list behavior to apply to all gates. This will cover 95% of use cases

there will likely be some checks that need "extra" exclusions that other gates don't need. There will also be some gates who "want to put X files back in". so here's the new pattern

repo-level exclude lists (all gates ignore these files)
optional gate-level exclude lists (any files or folders here are added to the exclude list *for this gate only*)
optional gate-level include list (add previously excluded files for specific checks) example - maybe we exlcude docs for all checks except string duplication?

### LINKED ISSUE #135: sm agent install --target copilot writes to project-local path instead of user-global ~/.copilot/skills/
### Current Behavior
`sm agent install --target copilot` writes skill files to the **project-local** `.copilot/skills/slopmop/` directory.

VS Code Copilot Chat loads user-level skills from `~/.copilot/skills/`, not from the workspace `.copilot/` directory. The installed SKILL.md is never picked up by Copilot because it is in the wrong location.

### Expected Behavior
`--target copilot` should install to `~/.copilot/skills/slopmop/SKILL.md` (the user-level skill directory that Copilot Chat actually reads from), or at minimum document the correct path so the user knows to copy it.

### Steps to Reproduce
1. Have an existing placeholder at `~/.copilot/skills/slopmop/SKILL.md`
2. Run `sm agent install --target copilot --force` from a project root
3. Observe files written to `<project>/.copilot/skills/slopmop/SKILL.md`
4. The user-level `~/.copilot/skills/slopmop/SKILL.md` remains unchanged
5. Copilot Chat still loads the empty placeholder, not the real skill content

### Additional Context
- Environment: macOS, VS Code Insiders, sm 0.13.0
- The `--target claude` install writes to `.claude/` (project-local) which is correct for Claude Code since it reads from the workspace.
- Copilot is different: it reads skills from the user-global `~/.copilot/skills/` path.
- Workaround: manually copy the installed SKILL.md to `~/.copilot/skills/slopmop/SKILL.md`
- Impact Level: medium (skill is installed but never loaded — silent failure)

ScienceIsNeato:
## Summary
- prevent ERROR: buff resolve requires PR number and thread id: sm buff resolve <pr_number> <thread_id> GitHub helper subprocesses from hanging on stdin
- degrade isolated ESLint harness config failures to warnings so bare ✨ scanning the code for slop to mop

ℹ  Remediation mode: auto-fix disabled (pass --auto-fix to enable, or run `sm refit --finish` to transition to maintenance mode)
✅ 🔧 myopia:vulnerability-blindness.py: passed (0.00s)
✅ 🔧 laziness:repeated-code: passed (0.00s)
✅ 🔧 myopia:ambiguity-mines.py: passed (0.00s)
✅ 🔧 laziness:dead-code.py: passed (0.00s)
✅ 🔬 myopia:string-duplication.py: passed (0.00s)
✅ 🔬 deceptiveness:bogus-tests.py: passed (0.00s)
✅ 🔧 laziness:sloppy-formatting.py: passed (0.00s)
✅ 🔧 overconfidence:missing-annotations.py: passed (0.00s)
✅ 🔧 overconfidence:type-blindness.py: passed (0.00s)
✅ 🔬 myopia:code-sprawl: passed (0.00s)
✅ 🔧 laziness:complexity-creep.py: passed (0.00s)
⚠️ 🔧 overconfidence:untested-code.py: warned (0.00s)
⏭️ overconfidence:coverage-gaps.py: skipped (0.00s)
✅ 🔬 deceptiveness:debugger-artifacts: passed (0.00s)
════════════════════════════════════════════════════════════
✨ NO SLOP DETECTED · 12 checks passed, 1 warned (🔧 8 foundation, 🔬 4 diagnostic)
   in 0.3s
   📦 13/14 from cache (commit 77ac44c, 8m to 8m ago)
   🔄 Fresh run: rerun `sm swab --no-cache` if you need uncached results.
════════════════════════════════════════════════════════════

⚠️  WARNINGS (non-blocking):
   • 🔧 overconfidence:untested-code.py
     └─ No project virtual environment found
     📄 full details: .slopmop/logs/overconfidence_untested-code.py.log
     💡 Create a venv so this check can run against your project:
  cd /Users/pacey/.codex/worktrees/4c56/slop-mop && python3 -m venv venv && source venv/bin/activate && pip install -e '.[dev]' stays usable on flat-config repos
- install Copilot repo instructions locally while sending the Copilot skill to the user home directory
- add shared repo-wide path exclusion handling from top-level config plus .gitignore, with per-gate extra excludes and include-back paths
- surface the new exclusion state in config/help output and cover the batch with regression tests

## Verification
- pytest tests/unit/test_agent_install.py tests/unit/test_sm_cli_parser.py tests/unit/test_javascript_eslint_expect.py tests/unit/test_buff_workflow_actions.py tests/unit/test_quality_checks.py tests/unit/test_security_checks.py tests/unit/test_gate_dodging_check.py tests/unit/test_sm_cli_config.py tests/unit/test_registry_runtime_path_filters.py tests/unit/test_sm_cli.py -q -k 'not TestMain'
- commit hook swab pass during git commit

Closes #95
Closes #99
Closes #100
Closes #101
Closes #135

### DIFF
--- a/slopmop/agent_install/installer.py
+++ b/slopmop/agent_install/installer.py
@@ -2,12 +2,17 @@
 
 from __future__ import annotations
 
+import os
 from dataclasses import dataclass, field
 from pathlib import Path
 from typing import List
 
 from slopmop.agent_install.loader import load_assets
-from slopmop.agent_install.registry import TARGETS, expand_target
+from slopmop.agent_install.registry import (
+    TARGETS,
+    expand_target,
+    uses_user_home_destination,
+)
 
 
 @dataclass
@@ -25,6 +30,7 @@ def install_agent_templates(
 ) -> InstallReport:
     """Install template files for the given target into *project_root*."""
     project_root = project_root.resolve()
+    user_home = Path(os.path.expanduser("~")).resolve()
     report = InstallReport(project_root=project_root)
 
     target_keys = expand_target(target)
@@ -38,7 +44,10 @@ def install_agent_templates(
             continue
 
         for asset in assets:
-            destination = project_root / asset.destination_relpath
+            if uses_user_home_destination(key, asset.destination_relpath):
+                destination = user_home / asset.destination_relpath
+            else:
+                destination = project_root / asset.destination_relpath
             try:
                 if destination.exists() and not force:
                     report.skipped.append(destination)
--- a/slopmop/agent_install/installer.py
+++ b/slopmop/agent_install/installer.py
@@ -12,6 +12,7 @@ from slopmop.agent_install.registry import (
     expand_target,
     uses_user_home_destination,
 )
+from slopmop.utils import posix_relpath_to_path
 
 
 @dataclass
@@ -53,9 +54,13 @@ def install_agent_templates(
                         "user home directory (HOME may be unset)"
                     )
                     continue
-                destination = user_home / asset.destination_relpath
+                destination = user_home / posix_relpath_to_path(
+                    asset.destination_relpath
+                )
             else:
-                destination = project_root / asset.destination_relpath
+                destination = project_root / posix_relpath_to_path(
+                    asset.destination_relpath
+                )
             try:
                 if destination.exists() and not force:
                     report.skipped.append(destination)
--- a/slopmop/agent_install/installer.py
+++ b/slopmop/agent_install/installer.py
@@ -2,7 +2,6 @@
 
 from __future__ import annotations
 
-import os
 from dataclasses import dataclass, field
 from pathlib import Path
 from typing import List
@@ -30,11 +29,10 @@ def install_agent_templates(
 ) -> InstallReport:
     """Install template files for the given target into *project_root*."""
     project_root = project_root.resolve()
-    _raw_home = os.path.expanduser("~")
-    if _raw_home.startswith("~") or not Path(_raw_home).is_absolute():
-        user_home: Path | None = None
-    else:
-        user_home = Path(_raw_home).resolve()
+    try:
+        user_home: Path | None = Path.home().resolve()
+    except RuntimeError:
+        user_home = None
     report = InstallReport(project_root=project_root)
 
     target_keys = expand_target(target)
--- a/slopmop/agent_install/installer.py
+++ b/slopmop/agent_install/installer.py
@@ -30,7 +30,11 @@ def install_agent_templates(
 ) -> InstallReport:
     """Install template files for the given target into *project_root*."""
     project_root = project_root.resolve()
-    user_home = Path(os.path.expanduser("~")).resolve()
+    _raw_home = os.path.expanduser("~")
+    if _raw_home.startswith("~") or not Path(_raw_home).is_absolute():
+        user_home: Path | None = None
+    else:
+        user_home = Path(_raw_home).resolve()
     report = InstallReport(project_root=project_root)
 
     target_keys = expand_target(target)
@@ -45,6 +49,12 @@ def install_agent_templates(
 
         for asset in assets:
             if uses_user_home_destination(key, asset.destination_relpath):
+                if user_home is None:
+                    report.errors.append(
+                        f"{asset.destination_relpath}: skipped — cannot resolve "
+                        "user home directory (HOME may be unset)"
+                    )
+                    continue
                 destination = user_home / asset.destination_relpath
             else:
                 destination = project_root / asset.destination_relpath
--- a/slopmop/agent_install/registry.py
+++ b/slopmop/agent_install/registry.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
-from pathlib import PurePosixPath
+from pathlib import Path
 from typing import Dict, List, Tuple
 
 
@@ -60,8 +60,9 @@ TARGETS: Dict[str, InstallTarget] = {
 }
 
 ALL_KEYS: Tuple[str, ...] = tuple(TARGETS.keys())
-INSTALL_HELP_PREVIEW_ROOT = PurePosixPath(".slopmop/tmp")
-INSTALL_HELP_HOME_PREVIEW_ROOT = PurePosixPath("~")
+INSTALL_HELP_PREVIEW_ROOT = Path(".slopmop") / "tmp"
+INSTALL_HELP_HOME_PREVIEW_ROOT = Path("~")
+COPILOT_SKILLS_PATH = ".copilot/skills/"
 
 ALIASES: Dict[str, List[str]] = {
     "all": list(ALL_KEYS),
@@ -84,13 +85,13 @@ def expand_target(target: str) -> List[str]:
 
 def uses_user_home_destination(target: str, destination_relpath: str) -> bool:
     """Return whether a template asset installs under the user's home dir."""
-    return target == "copilot" and destination_relpath.startswith(".copilot/skills/")
+    return target == "copilot" and destination_relpath.startswith(COPILOT_SKILLS_PATH)
 
 
 def preview_install_paths(
     target: str,
-    preview_root: PurePosixPath = INSTALL_HELP_PREVIEW_ROOT,
-    home_preview_root: PurePosixPath = INSTALL_HELP_HOME_PREVIEW_ROOT,
+    preview_root: Path = INSTALL_HELP_PREVIEW_ROOT,
+    home_preview_root: Path = INSTALL_HELP_HOME_PREVIEW_ROOT,
 ) -> List[str]:
     """Return preview install destinations for one target.
 
@@ -104,7 +105,7 @@ def preview_install_paths(
 
     paths: List[str] = []
     for asset in load_assets(TARGETS[target].template_dir):
-        relpath = PurePosixPath(asset.destination_relpath)
+        relpath = Path(asset.destination_relpath)
         if uses_user_home_destination(target, asset.destination_relpath):
             paths.append(str(home_preview_root / relpath))
         else:
--- a/slopmop/agent_install/registry.py
+++ b/slopmop/agent_install/registry.py
@@ -61,6 +61,7 @@ TARGETS: Dict[str, InstallTarget] = {
 
 ALL_KEYS: Tuple[str, ...] = tuple(TARGETS.keys())
 INSTALL_HELP_PREVIEW_ROOT = PurePosixPath(".slopmop/tmp")
+INSTALL_HELP_HOME_PREVIEW_ROOT = PurePosixPath("~")
 
 ALIASES: Dict[str, List[str]] = {
     "all": list(ALL_KEYS),
@@ -81,9 +82,15 @@ def expand_target(target: str) -> List[str]:
     raise ValueError(f"Unknown agent install target: {target}")
 
 
+def uses_user_home_destination(target: str, destination_relpath: str) -> bool:
+    """Return whether a template asset installs under the user's home dir."""
+    return target == "copilot" and destination_relpath.startswith(".copilot/skills/")
+
+
 def preview_install_paths(
     target: str,
     preview_root: PurePosixPath = INSTALL_HELP_PREVIEW_ROOT,
+    home_preview_root: PurePosixPath = INSTALL_HELP_HOME_PREVIEW_ROOT,
 ) -> List[str]:
     """Return preview install destinations for one target.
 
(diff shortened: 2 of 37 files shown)


ScienceIsNeato:
[fixed_in_code] Fixed in commit a05a1f6. Removed `latest_completed_for_head` from the `_WorkflowRunState` TypedDict and the corresponding computation block in `_workflow_run_state`. The field was computed but never read by `_resolve_fresh_run`, which only uses `latest_for_head`.

ScienceIsNeato:
[fixed_in_code] Fixed in commit a05a1f6. Added `dedupe_str_list` to `slopmop/utils/__init__.py` and retired both `_dedupe_str_list` (sm.py) and `_dedupe_strings` (registry.py) in favour of the shared implementation. Both modules now import from utils.

ScienceIsNeato:
[fixed_in_code] Fixed in commit a05a1f6. `install_agent_templates` now validates that `expanduser('~')` actually expanded before using the result. If the expanded path still starts with `~` or is not absolute, `user_home` is set to `None`. Any asset that requires a user-home destination records an error in the report and is skipped.

ScienceIsNeato:
[fixed_in_code] Fixed in commit a05a1f6. `_load_gitignore_exclude_paths` now wraps `read_text` in a try/except for `UnicodeDecodeError` and returns `[]` with a `logger.warning` instead of propagating the exception. A non-UTF-8 .gitignore no longer crashes sm commands.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 2ed6ece. `_simple_path_filters` now skips any value that contains `/`, since nested paths like `vendor/generated` cannot safely flow into `exclude_dirs`-style fields on checks that use the old `part in excluded_set` pattern. Those values still reach the right checks via `_path_pattern_variants` / `exclude_patterns`.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 2ed6ece. Removed `token == 'head'` from the git-SHA false-positive heuristic. The remaining checks (`is_placeholder_sha(`, `rev-parse`, `_current_head`, `token == 'sha'`, `token.endswith('_sha')`) are specific enough to suppress git-SHA false positives without risking suppression of real secrets that happen to appear near an unrelated `head` variable.

ScienceIsNeato:
[fixed_in_code] Commit ec60cdb adds _cached_check_schema_fields() decorated with @lru_cache(maxsize=None) in registry.py. _merge_runtime_path_filters now calls this helper instead of check_class({}).get_full_config_schema() directly, so the throwaway instance is only created once per class per process lifetime.

ScienceIsNeato:
[fixed_in_code] Commit ec60cdb moves the getattr(jinja2_module, ...) calls inside the try block and changes the except clause to catch (ImportError, AttributeError). A broken or stub jinja2 install that raises AttributeError on attribute lookup now produces the friendly 'Jinja2 not installed' skip result instead of propagating an unhandled exception.

slopmop/agent_install/installer.py
```diff
@@ -25,6 +30,11 @@ def install_agent_templates(
 ) -> InstallReport:
     """Install template files for the given target into *project_root*."""
     project_root = project_root.resolve()
+    _raw_home = os.path.expanduser("~")
```
ScienceIsNeato: this looks like something that should be abstracted by os

slopmop/agent_install/registry.py
```diff
@@ -81,9 +82,15 @@ def expand_target(target: str) -> List[str]:
     raise ValueError(f"Unknown agent install target: {target}")
 
 
+def uses_user_home_destination(target: str, destination_relpath: str) -> bool:
+    """Return whether a template asset installs under the user's home dir."""
+    return target == "copilot" and destination_relpath.startswith(".copilot/skills/")
```
ScienceIsNeato: that should be a constant representing COPILOT_SKILLS_PATH, not a string literal

slopmop/agent_install/registry.py
```diff
@@ -95,7 +102,11 @@ def preview_install_paths(
...
         return []
 
-    return [
-        str(preview_root / PurePosixPath(asset.destination_relpath))
-        for asset in load_assets(TARGETS[target].template_dir)
-    ]
+    paths: List[str] = []
+    for asset in load_assets(TARGETS[target].template_dir):
+        relpath = PurePosixPath(asset.destination_relpath)
+        if uses_user_home_destination(target, asset.destination_relpath):
```
ScienceIsNeato, line 109: all of the stuff in this fiule looks 'nix specific, so the windows side of this is going to fail. need to abstract inastead of "/" to use "get_path_separator")

ScienceIsNeato:
[fixed_in_code] Commit a908995 replaces PurePosixPath with pathlib.Path throughout registry.py. INSTALL_HELP_PREVIEW_ROOT and INSTALL_HELP_HOME_PREVIEW_ROOT now use Path, and preview_install_paths uses Path for relpath construction - so path separators follow the OS.

ScienceIsNeato:
[fixed_in_code] Commit a908995 replaces os.path.expanduser("~") + manual validation with try: user_home = Path.home().resolve() / except RuntimeError: user_home = None. Path.home() is the stdlib-idiomatic cross-platform way to get the home dir; import os removed.

ScienceIsNeato:
[fixed_in_code] Commit a908995 extracts COPILOT_SKILLS_PATH = ".copilot/skills/" as a module-level constant in registry.py. uses_user_home_destination() now references the constant instead of the inline string literal.

slopmop/agent_install/installer.py
```diff
@@ -38,7 +46,16 @@ def install_agent_templates(
...
         for asset in assets:
-            destination = project_root / asset.destination_relpath
+            if uses_user_home_destination(key, asset.destination_relpath):
+                if user_home is None:
+                    report.errors.append(
+                        f"{asset.destination_relpath}: skipped — cannot resolve "
+                        "user home directory (HOME may be unset)"
+                    )
+                    continue
+                destination = user_home / asset.destination_relpath
```
ScienceIsNeato: this is another example of using hard coded platform dependedt separators. search this whole space for forward slashes and replaced them all with calls to a get separator helper utility

ScienceIsNeato:
[fixed_in_code] Commit 0afaef7 reverts INSTALL_HELP_PREVIEW_ROOT and INSTALL_HELP_HOME_PREVIEW_ROOT back to PurePosixPath. These are display-only constants for CLI help text - PurePosixPath guarantees forward slashes regardless of platform, which was the original intent. The installer uses os-native pathlib.Path for actual file operations.

ScienceIsNeato:
[fixed_in_code] Commit 0afaef7 adds _glob_match() in utils/__init__.py. When a pattern contains **, it splits on **/ and builds a regex where each **/ becomes (?:[^/]+/)* (zero or more path components). is_path_excluded now calls _glob_match instead of fnmatch for wildcard patterns, so **/*.snap correctly matches bar.snap at root as well as foo/bar/baz.snap in subdirectories.

ScienceIsNeato:
[fixed_in_code] Commit 2f80ea5 rewrites _glob_match() with a char-by-char translation instead of split('**/').  ** is now always translated to .* (matches anything including /) and **/ to (?:.+/)? (optional directory prefix). The old split() approach left trailing ** as two un-joined [^/]* fragments - vendor/** now correctly matches vendor/foo, vendor/foo/bar, and vendor/foo/bar/baz.

ScienceIsNeato:
[fixed_in_code] Commit 2f80ea5 adds posix_relpath_to_path() helper in slopmop/utils/__init__.py that converts a posix-style destination_relpath string to a native OS Path. installer.py now calls this helper at both join sites (user_home and project_root) instead of passing the posix string directly to pathlib's / operator. Searched full agent_install space - COPILOT_SKILLS_PATH and the INSTALL_HELP_PREVIEW_ROOT constants use PurePosixPath intentionally (display-only, always forward slashes).

ScienceIsNeato:
[fixed_in_code] Commit 83b2c88 adds a [ branch to the _glob_match char-by-char loop. When [ is encountered it scans forward to the matching ] (respecting ] as a literal immediately after the opening bracket, and ! / ^ negation) and appends the full character class verbatim to the regex - no re.escape applied. Patterns like **/[Bb]uild/**, **/*.[ch] now match correctly. Patterns without ** still delegate to fnmatch which already handles [] fine.

PR #146, opened 2026-04-24
Fix sm_env doctor and prepare-release workflow

ScienceIsNeato:
Summary of changes:
- Fixed sm_env doctor issues.
- Cleaned up the prepare-release workflow.
- Improved release.sh error handling.
- Enabled Actions PR creation via repository settings, allowing the workflow to use the default github.token for PR operations.

### DIFF
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -33,20 +33,11 @@ jobs:
     name: Prepare Release (${{ inputs.bump }})
     runs-on: ubuntu-latest
     steps:
-      - name: Verify release token configured
-        env:
-          RELEASE_TOKEN: ${{ secrets.BUMP_VERSION_TOKEN }}
-        run: |
-          if [[ -z "$RELEASE_TOKEN" ]]; then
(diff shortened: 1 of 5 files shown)


ScienceIsNeato:
[fixed_in_code] Removed the unreachable _warn fallback in slopmop/doctor/sm_env.py; _build_report now returns FAIL directly because run() only calls it when one of missing, validator_rejects, or version_violations is present. Fixed in ebaf1f5.

ScienceIsNeato:
[fixed_in_code] Updated scripts/release.sh to stop pointing at BUMP_VERSION_TOKEN and instead direct users to enable Actions PR creation or rerun with a GH_TOKEN that has pull-request write access. The repo setting now allows Actions to create PRs, and the workflow uses github.token again. Fixed in ebaf1f5.

ScienceIsNeato:
[fixed_in_code] Updated scripts/release.sh so the failure guidance now matches the actual workflow and repo configuration: it references enabling GitHub Actions PR creation or using GH_TOKEN with pull-request write access, not a nonexistent workflow secret. Fixed in ebaf1f5.

PR #147, opened 2026-04-24
fix: make release reruns idempotent

ScienceIsNeato:
- reuse existing remote release branches during reruns
- add regression coverage for release.sh rerun behavior

### DIFF
--- a/.github/workflows/slopmop.yml
+++ b/.github/workflows/slopmop.yml
@@ -35,7 +35,7 @@ jobs:
         run: pip install -e ".[all]"
 
       - name: Run validator and doctor unit tests
-        run: python -m pytest tests/unit/test_subprocess_validator.py tests/unit/test_doctor.py tests/unit/test_doctor_checks.py -v
+        run: python -m pytest tests/unit/test_subprocess_validator.py tests/unit/test_doctor.py tests/unit/test_doctor_checks.py tests/unit/test_release_script.py -v
 
   dogfood:
     name: Final Dogfood Sanity Check (blocking)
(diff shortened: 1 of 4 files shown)


ScienceIsNeato:
[invalid_with_explanation] I investigated this before changing code. On this repo, `gh pr list --head definitely-no-such-release-branch --base main --state open --json url --jq '.[0].url'` returns an empty string with exit code 0, not `null`, so `[[ -n "$EXISTING_PR_URL" ]]` does not misfire here. I left the code unchanged rather than add a churn-only patch.

ScienceIsNeato:
[fixed_in_code] Fixed in commit bbfbb5d. Added a second regression in tests/unit/test_release_script.py that simulates an existing open release PR, asserts the script reuses that URL, and verifies it does not invoke `gh pr create` again.

ScienceIsNeato:
[fixed_in_code] Fixed in commit bbfbb5d. Updated .github/workflows/slopmop.yml so CI's explicit unit-test list now includes tests/unit/test_release_script.py, keeping the release rerun regression under continuous coverage.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 27f82f6. I verified the no-match behavior locally, then made the release PR lookup explicitly empty-safe anyway by changing the filter to `.[0].url // empty`, so reruns cannot treat a missing PR as a truthy string.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 27f82f6. Removed the unused `log_path` parameter from the fake-gh helper so the helper signature now matches how the test harness actually wires logging through `GH_LOG`.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 27f82f6. Removed the unused `log_path` parameter from the fake-gh helper so the helper signature now matches how the test harness actually wires logging through `GH_LOG`.

PR #149, opened 2026-04-24
fix: repair timing history and stale gate applicability

ScienceIsNeato:
## Summary
- ignore cached and failed timing samples so ETA history only learns from real runs
- stop requirements-only repos from enabling Python gates when there is no Python source
- repair stale built-in gate applicability on upgrade instead of telling users to rerun init

## Validation
- pytest tests/unit/test_display_timings.py -q
- pytest tests/unit/test_detection.py tests/unit/test_base_check.py tests/unit/test_config_debt_check.py tests/unit/test_python_checks.py -q
- pytest tests/unit/test_upgrade_migrations.py -q
- ./sm swab

## Context
These changes came out of live dogfooding against fogofdog-frontend, where stale applicability and timing-history behavior were creating bad guidance.

### DIFF
--- a/STATUS.md
+++ b/STATUS.md
@@ -1,5 +1,30 @@
 # Project Status
 
+## 2026-04-24 Delta: requirements-only repos no longer activate Python gates
+
+Branch: `friction`
+
+**Work completed:**
+- Tightened Python project heuristics so `requirements.txt` alone no longer
+  marks a repo as Python for slop-mop detection and applicability.
+- Aligned three surfaces to use the same rule:
+  - `slopmop/cli/detection.py` for `sm init`
+  - `slopmop/checks/mixins.py` for Python gate applicability
+  - `slopmop/checks/quality/config_debt.py` for `silenced-gates`
+- Added regressions so requirements-only repos without `.py` files:
+  - do not detect as Python during init
+  - do not make Python gates applicable at runtime
+  - do not trigger stale Python disabled-gate warnings
+- Updated Python-check tests that previously used `requirements.txt` as the
+  only Python-project marker.
+
+**Validation:**
+- `pytest tests/unit/test_detection.py tests/unit/test_base_check.py tests/unit/test_config_debt_check.py tests/unit/test_python_checks.py -q` ✅
+- `./sm swab` ✅ (17/17 checks passed)
+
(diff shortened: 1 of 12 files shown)


ScienceIsNeato:
[fixed_in_code] Fixed in b1a7619. looks_like_python_project() no longer has the dead requirements-only tail; it now returns on strong manifests or a single has_python_source_files() probe.

ScienceIsNeato:
[fixed_in_code] Fixed in b1a7619. _detect_python() no longer delegates to looks_like_python_project(); it stays manifest-first and only uses a bounded excluded-dir scan when requirements.txt is the lone Python signal.

ScienceIsNeato:
[fixed_in_code] Fixed in b1a7619. has_python_source_files() now uses os.walk with directory pruning and early exit instead of an unbounded Path.rglob('*.py').

ScienceIsNeato:
[fixed_in_code] Fixed in b1a7619. _has_python_markers() no longer routes through the broad project helper; it uses strong manifests first and a bounded excluded-dir scan only for the requirements.txt edge case.

ScienceIsNeato:
[fixed_in_code] Fixed in b1a7619. looks_like_python_project() now calls has_python_source_files() once, so the redundant traversal is gone.

ScienceIsNeato:
[fixed_in_code] Fixed in ffb48f6. detection.py now starts with the module docstring, keeps a single canonical mixins import block below it, and imports has_python_source_files from that block so the helper remains available at runtime.

ScienceIsNeato:
[fixed_in_code] Fixed in ffb48f6. config_debt.py now relies on has_python_source_files() defaults for excluded directories and only overrides max_depth, so the copied exclusion set is gone.

ScienceIsNeato:
[fixed_in_code] Fixed in a375b97. detection.py now aliases _DETECTION_EXCLUDED_DIRS to the shared _PYTHON_SOURCE_EXCLUDE_DIRS set from mixins.py, so detection keeps its local name without carrying a second copy.

ScienceIsNeato:
[fixed_in_code] Fixed in c5a9c4e. has_python_source_files() now distinguishes None from set(), so callers can pass an explicit empty exclude_dirs set to disable exclusions instead of silently falling back to the defaults.

PR #159, opened 2026-04-30
docs: polish PyPI README for 1.0.1

ScienceIsNeato:
- Add human-readable sm swab figure and screenshot asset.
- Make refit the first-class onboarding path, with baseline as temporary fallback.
- Include sm status in the main verb table and simplify CI guidance.
- Use durable GitHub doc links with descriptive labels.
- Bump version to 1.0.1 for the PyPI long-description patch.

Validation:
- ./sm swab --no-cache --output-file .slopmop/last_swab.json
- ./sm scour --no-cache --output-file .slopmop/last_scour.json


ScienceIsNeato:
[fixed_in_code] Fixed locally. Updated the README figure caption to use explicit HTML <code> tags for sm swab and --porcelain inside the figcaption, so GitHub/PyPI render inline code correctly.

slopmop/reporting/timings.py
```diff
@@ -45,6 +45,15 @@
 # timing file from accumulating obsolete data.
 MAX_AGE_DAYS = 30
 
+# Cached or otherwise short-circuited runs should not be recorded as timing
+# samples, but older versions or interrupted runs may have already polluted
+# timing history. If a new real run is dramatically slower than every stored
+# sample, the old history is more harmful than helpful for ETA/budgeting.
+_CACHE_POISON_MIN_DURATION = 5.0
```
ScienceIsNeato: 5 what? seconds? that seems random.

slopmop/reporting/timings.py
```diff
@@ -45,6 +45,15 @@
...
+_CACHE_POISON_MIN_GAP = 5.0
+_CACHE_POISON_RATIO = 3.0
```
ScienceIsNeato: make this 10

slopmop/reporting/timings.py
```diff
@@ -302,6 +311,32 @@ def _migrate_v1_entry(entry: Dict[str, Any]) -> Dict[str, Any]:
     }
 
 
+def _should_reset_implausibly_fast_history(
```
ScienceIsNeato, line 317: this should be done once during an upgrade, not as a matter of course all the time. that's chasing a weird last battle

README.md
```diff
@@ -183,31 +197,33 @@ sm config --disable laziness:complexity-creep.py
 ```
 
 Disabling a gate should be temporary. If a gate is wrong, tune it or file the
-tooling bug. If the repo is not ready yet, use refit or baseline mode instead of
-pretending the problem is gone.
+tooling bug. If the repo is not ready yet, use refit first. Use baseline mode
+only as a short-term unblocker when refit is not practical in the moment.
 
-Migration behavior is documented in [DOCS/MIGRATIONS.md](https://github.com/ScienceIsNeato/slop-mop/blob/main/DOCS/MIGRATIONS.md).
```
ScienceIsNeato, line 189: there are lots of changes acorss this PR that change filename links to general descripts. that's not what we wanted . we wanted to fix the underlying links. please undo all these changes in the readme

ScienceIsNeato:
[fixed_in_code] Fixed in 3d457ba. Reverted all link texts to filename format (DOCS/FOO.md) while keeping the absolute GitHub URLs. Descriptive text removed throughout.

ScienceIsNeato:
[fixed_in_code] Fixed in 3d457ba. Added inline unit comments to all four constants: _CACHE_POISON_MIN_DURATION = 5.0  # seconds, _CACHE_POISON_MIN_GAP = 5.0  # seconds, _CACHE_POISON_RATIO = 10.0, _CACHE_POISON_FAST_HISTORY_MAX = 10.0  # seconds

ScienceIsNeato:
[fixed_in_code] Fixed in 3d457ba. Raised _CACHE_POISON_RATIO from 3.0 to 10.0 per request.

ScienceIsNeato:
[fixed_in_code] Fixed in 3d457ba. The reset now fires at most once per entry. When it fires, cache_poison_reset_applied=True is persisted to timings.json. On subsequent runs we skip the check entirely if the marker is set, so it can't keep resetting on every slow run.

ScienceIsNeato:
[invalid_with_explanation] Bugbot conflated two separate review threads. The 'make this 10' comment (thread PRRT_kwDORBxXu85-ri6z) was on _CACHE_POISON_RATIO at line 54 — that was changed from 3.0 to 10.0 in 3d457ba. The comment at line 52 was 'what? seconds? that seems random' — asking for units, not a value change. That was also addressed in 3d457ba by adding inline unit comments. Both threads are resolved.

slopmop/reporting/timings.py
```diff
@@ -45,6 +45,15 @@
 # timing file from accumulating obsolete data.
 MAX_AGE_DAYS = 30
 
+# Cached or otherwise short-circuited runs should not be recorded as timing
+# samples, but older versions or interrupted runs may have already polluted
+# timing history. If a new real run is dramatically slower than every stored
+# sample, the old history is more harmful than helpful for ETA/budgeting.
+_CACHE_POISON_MIN_DURATION = 5.0  # seconds — real run must be at least this long
```
ScienceIsNeato: this isn't the way to do this. the way to check to see if cached runs are in the list is to basically see if the if there are outliers in the dataset, not a specific number. most tests don't take 5 seconds to begin with

slopmop/reporting/timings.py
```diff
@@ -45,6 +45,15 @@
...
+_CACHE_POISON_MIN_GAP = 5.0  # seconds — gap above historical max
```
ScienceIsNeato: you can't use literal seconds here - must be a ratio of some kind

slopmop/reporting/timings.py
```diff
@@ -302,6 +311,32 @@ def _migrate_v1_entry(entry: Dict[str, Any]) -> Dict[str, Any]:
...
 
 
+def _should_reset_implausibly_fast_history(
+    samples: List[float],
+    new_duration: float,
+) -> bool:
+    """Return True when old samples look like cached/short-circuit timings.
+
+    A single slow run can be noise, so the threshold is deliberately blunt:
+    the previous history must be entirely sub-10s, and the new real duration
```
ScienceIsNeato: this is all sounding like it is based on heuristics, but it is based on nothing., take another swing on this and ground it in mathematics.

ScienceIsNeato:
[fixed_in_code] Reworked in 938e03f. The new `_should_reset_implausibly_fast_history` drops all absolute-second thresholds entirely and uses a single scale-invariant criterion: `new_duration / max(history) >= 10`. This works equally well for sub-second checks and multi-minute checks — no hardcoded floor required. Also requires ≥2 existing samples so a single noisy data point can't trigger a spurious reset.

ScienceIsNeato:
[fixed_in_code] Fixed in 938e03f. Removed `_CACHE_POISON_MIN_GAP` (absolute seconds) along with `_CACHE_POISON_MIN_DURATION` and `_CACHE_POISON_FAST_HISTORY_MAX`. The only remaining constant is the dimensionless `_CACHE_POISON_RATIO = 10.0`, and the check is purely `new_duration / max(history) >= 10`.

ScienceIsNeato:
[fixed_in_code] Rewritten in 938e03f with an explicit mathematical basis. The docstring now grounds the criterion in log-normal timing distributions: requiring `new / max(history) >= 10` means the two populations must differ by at least one order of magnitude, which under a log-normal model makes it statistically implausible they share the same underlying process. The three heuristic constants (min duration, min gap, max fast history) are gone — just the dimensionless ratio.

PR #181, opened 2026-05-05
ci: allow manual PyPI release dispatch

ScienceIsNeato:
## Summary

- Add manual workflow_dispatch support to Release to PyPI.
- Require manual release dispatches to run from main.
- Add an optional version input that must match pyproject.toml.
- Document the Prepare Release -> merge release PR -> Release to PyPI flow.

## Validation

- ./sm swab --static
- ./sm scour --output-file .slopmop/last_scour.json


ScienceIsNeato:
[fixed_in_code] Fixed in commit 3998840. Manual publish dispatch now validates the full ref is refs/heads/main in addition to ref_name, so a tag named main cannot satisfy the branch guard.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 3998840. The release docs now state that merge-to-main auto-publishes on pyproject.toml changes, while action=publish is the manual recovery/double-check path. The old prepare-release workflow was removed, so there is one Actions entry.

PR #184, opened 2026-05-06
ci: make release workflow single-click

ScienceIsNeato:
## Summary

- update GitHub-owned workflow actions to Node 24-compatible major versions
- make the manual Release workflow bump, commit, validate, build, publish to PyPI, and create the GitHub release in one run
- keep the pyproject.toml push trigger for the PR-based release script fallback while using [skip release] on workflow-created bump commits to avoid duplicate runs

## Validation

- bash -n scripts/release.sh
- ./sm swab --static
- ./sm scour --output-file .slopmop/last_scour.json
- VS Code diagnostics: clean for release.yml, docs, release.sh, STATUS.md
- actionlint not run locally; actionlint is not installed

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -39,6 +39,9 @@ on:
           - minor
           - major
 
+env:
+  FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true
+
 # Prevent concurrent releases, but never cancel an in-progress publish. PyPI
 # forbids re-uploading the same version, so a partial publish caused by
 # cancellation would be permanently broken.
@@ -60,7 +63,7 @@ jobs:
       version: ${{ steps.resolve.outputs.version }}
       previous_version: ${{ steps.resolve.outputs.previous_version }}
       tag: ${{ steps.resolve.outputs.tag }}
-      commit_sha: ${{ steps.commit.outputs.commit_sha || github.sha }}
+      commit_sha: ${{ steps.resolve.outputs.commit_sha || steps.commit.outputs.commit_sha || github.sha }}
     steps:
       - uses: actions/checkout@v5
         with:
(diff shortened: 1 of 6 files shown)


ScienceIsNeato:
[fixed_in_code] Fixed in commit f77dae0. Added an explicit manual-dispatch sync step before the version bump is computed and committed: the Release workflow now runs git fetch origin main followed by git pull --ff-only origin main before editing pyproject.toml, so stale checkouts fail fast before creating the release commit.

ScienceIsNeato:
[fixed_in_code] Fixed in commit f77dae0. Removed the duplicate ./sm scour --output-file .slopmop/last_scour.json validation bullet from STATUS.md.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 398a964. Added FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true at workflow scope for Release, slopmop, and SARIF workflows so artifact actions execute on Node 24 even while upload-artifact/download-artifact v5 remain Node 20-backed by default.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 398a964. Manual release bump commits now include [release-run:${GITHUB_RUN_ID}], and rerun attempts look up that original commit on origin/main and reuse its version/commit_sha instead of applying the bump again.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 7eed17d. Removed upload-artifact/download-artifact from the Release workflow entirely: publish-pypi and GitHub Release now checkout the release commit and rebuild dist locally, and the SARIF workflow no longer uses diagnostic artifact uploads. FORCE_JAVASCRIPT_ACTIONS_TO_NODE24 no longer forces unsupported artifact actions.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 97782d5. The rerun lookup now uses git log --fixed-strings --grep for the [release-run:${GITHUB_RUN_ID}] marker, so the marker is matched literally instead of as a regex character class.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 97782d5. Rerun attempts now only reuse a prior release commit when one is found; if the first attempt failed before the bump commit existed, the rerun falls back to creating the bump normally.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 97782d5. Added contents: read to publish-pypi permissions so actions/checkout can read the release commit while the job still has id-token: write for trusted PyPI publishing.

PR #202, opened 2026-05-20
feat: sm gang — system-wide command intercepts + deep hooks + barnacle sweep

ScienceIsNeato:
## What's in this PR

Three new features and a barnacle sweep. **sm gang is the headliner.**

---

## sm gang — press-gang your shell

Commands don't volunteer. They're seized.

`sm gang install` wires **25 shell function intercepts** into your rc file machine-wide. Any time an agent (or you) reaches for a raw quality tool — `pytest`, `mypy`, `gh run view`, `black`, `npx knip` — the intercept fires instead:

```
[slop-mop] ⛔  pytest blocked
  Stop. File a barnacle: sm barnacle
  Tell the human what you were doing and wait.
```

No redirect hints. No suggested equivalent. The agent must escalate to a human.

### Why barnacle-only?

A redirect hint lets an agent self-heal. If slop-mop is blocking `pytest`, there's a reason — probably a gate is failing or the workflow isn't set up right. Showing the agent `pytest → sm swab` just teaches it to route around the problem. The barnacle-only design forces explicit escalation: the agent files an issue, stops, and waits. Humans see exactly what was attempted and why it failed.

### What gets intercepted

| Category | Commands |
|----------|----------|
| Test runners | `pytest`, `pytest --cov` |
| Formatters / linters | `black`, `isort`, `flake8`, `mypy`, `pyright` |
| Security scanners | `bandit`, `semgrep`, `pip-audit` |
| Dead code | `vulture`, `radon`, `diff-cover` |
| CI triage | `gh run list`, `gh run watch`, `gh run view`, `gh pr checks` |
| JS/TS tooling | `npx knip`, `npx eslint`, `npx prettier`, `npx tsc` |

`gh` itself is NOT blocked — only the CI-watching subcommands. `git` itself is NOT blocked — only the bypass flags (see deep hooks below).

### How it wires in

```bash
sm gang install --confirm "I understand this aliases commands system-wide on this machine"
```

Writes `~/.slopmop/aliases.sh` (the function intercepts) and appends a marker block to your rc file:

```bash
# BEGIN sm-gang
alias git="$HOME/.slopmop/bin/git_wrapper.sh"
source "$HOME/.slopmop/aliases.sh"
# END sm-gang
```

Idempotent — running install twice writes nothing if the content hasn't changed (SHA256-guarded). `sm gang uninstall` strips the block and leaves your rc file byte-for-byte identical to before. Fish shell gets a warning; bash and zsh are fully supported.

---

## sm commit-hooks install --deep — the git wrapper

The git wrapper (`git_wrapper.sh`) is now packaged as slopmop data and deployed to `~/.slopmop/bin/` by `sm gang install` (or standalone via `sm commit-hooks install --deep`).

It blocks every bypass spelling of `git commit --no-verify`:

- `--no-verify`, `-n`, `--no-validate`, `--no-hooks`, `--skip-checks`
- Environment variable bypasses: `SKIP=...`, `PRE_COMMIT_ALLOW_NO_CONFIG`

Scope is limited to hook-invoking git operations (commit, merge, rebase, cherry-pick, revert, am). All other git usage passes through untouched. GUI clients, other machine users, and non-hook commands are unaffected.

---

## laziness:dead-code.js gate

Knip-based dead code detection for JS/TS projects:

- Unused exports
- Unreferenced files
- Dead type declarations

Skips automatically on non-JS/TS repos. Zero config required for projects that already have a `package.json`.

---

## Barnacle sweep

All 7 barnacles filed during integration testing on `welcome-to-willville` are resolved and will auto-close on merge:

| # | Root cause | Fix |
|---|-----------|-----|
| #204 | `sm buff watch` exited silently when `gh` CLI not installed | `_run_gh` raises `TriageError` with install guidance |
| #205 | `sm buff inspect` didn't show failing gate output | Surfaces CI log content from artifact |
| #206 | Formatting gate ran on gitignored cursor-rules/ Python | `pyproject.toml` alone no longer triggers Python gates |
| #207 | pip-audit scanned slopmop's packages on a TS project | `VIRTUAL_ENV` must be inside project root to count as project env |
| #208 | detect-secrets false-positives on Cloudflare account IDs | Added public infra identifier keys to allowlist |
| #209 | `pip install slopmop` on ubuntu-latest loses `sm` from PATH | Workflow template now shows venv pattern |
| #210 | Remediation banner printed in CI after `sm refit --finish` | Banner suppressed when `CI` env var is set |

---

## Test plan

- [x] All 31 swab gates pass — 6/6 CI checks green
- [x] `sm gang install` without `--confirm` prints explanation, exits 1
- [x] `sm gang install --confirm "wrong phrase"` exits 1
- [x] Fish shell triggers warning, does not attempt rc wiring
- [x] Install → uninstall round-trip leaves rc file byte-for-byte identical
- [x] `sm gang install` twice: no redundant writes (content-hash guarded)
- [x] All 25 intercepted commands trigger barnacle-only block message
- [x] `gh` passthrough works for non-CI-watching subcommands
- [x] `git commit --no-verify` blocked; `git status` passes through
- [x] pip-audit skips on TS projects (no false CVEs from slopmop's env)


ScienceIsNeato:
[fixed_in_code] Fixed in commit 4c414b4. git_wrapper.sh now uses `[[ "$arg" =~ ^-[^-]*n ]]` regex to detect -n in any short-option cluster (e.g. -nm, -fn, -fnm) for commit/merge/cherry-pick commands. Standalone -n is also covered by the same regex.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 4c414b4. cmd_commit_hooks now runs _deep_hooks_install independently of _hooks_install — removed the `if result == 0` gate. Both results are combined via `max(result, deep_result)` so the machine-wide wrapper installs even when a foreign pre-commit hook blocks the repo hook.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 4c414b4. _hooks_status no longer returns early when the hooks directory doesn't exist — the early-return block was converted to an if/else so the deep hooks section always renders regardless of repo hook state.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 4c414b4. _deep_hooks_uninstall now calls content.find(DEEP_HOOK_END_MARKER, start_idx) after finding the start marker. If end_idx == -1 (end marker missing or corrupted), it prints a warning and skips the file, leaving it unchanged — no data is lost.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 4c414b4. The chmod call was moved out of the if/else branches and runs unconditionally after the content-comparison block, so the executable bit is always ensured even on the 'already up to date' path. File I/O is now wrapped in OSError handlers that print a clear message and return 1.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 4c414b4. Added _rc_has_marker() helper that reads rc files with encoding='utf-8', errors='replace' and catches OSError. Used in both _hooks_status and anywhere else the marker is checked, so a non-UTF-8 rc file or permissions error no longer breaks `sm commit-hooks status`.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 4c414b4. cmd_commit_hooks now short-circuits before the git-repo gate: if there is no hooks_dir AND --deep is set AND the action is 'uninstall', it calls _deep_hooks_uninstall() directly and returns. The 'Not a git repository' error only fires for operations that actually need the repo hooks dir.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 4c414b4. main() in git_wrapper.sh now scans all argv tokens to find the first non-option argument as the subcommand before any flag checking. Global git options (--no-pager, -c key=val) are skipped by the initial loop, so git_command is always the actual subcommand and the -n block fires correctly for e.g. `git --no-pager commit -n`.

ScienceIsNeato:
[fixed_in_code] Fixed in fc2ca4d — -n block now only applies to commit subcommand, not merge/cherry-pick

ScienceIsNeato:
[fixed_in_code] Fixed in fc2ca4d — added handling for combined -m<message> syntax to avoid regex false positives

ScienceIsNeato:
[fixed_in_code] Fixed in fc2ca4d — reversed uninstall order to clean RC files before deleting wrapper

ScienceIsNeato:
[fixed_in_code] Fixed in fc2ca4d — subcommand loop now skips git global option values (-c, -C, -x)

ScienceIsNeato:
[fixed_in_code] Fixed in ef82d5f — _deep_hooks_install now uses _rc_has_marker(rc_file) as a drop-in replacement

ScienceIsNeato:
[fixed_in_code] Fixed in ef82d5f — rc_file.read_text() now uses encoding='utf-8', errors='replace' with OSError handling

ScienceIsNeato:
[fixed_in_code] Fixed in ef82d5f — uninstall branch now uses max(result, deep_result) matching install branch behavior

ScienceIsNeato:
[fixed_in_code] Fixed in ef82d5f — _hooks_status now calls _deep_rc_candidates() and _rc_has_marker() instead of inline list

ScienceIsNeato:
[fixed_in_code] Fixed in commit d940b38. SKIP and PRE_COMMIT_ALLOW_NO_CONFIG checks are now gated behind a hook_commands guard — they only fire when git_command is one that invokes pre-commit/commit-msg hooks (commit, merge, rebase, cherry-pick, revert, am). Commands like 'git status' or 'git log' pass through cleanly even when SKIP is set. Logic extracted into check_hook_env_bypasses() helper.

ScienceIsNeato:
[fixed_in_code] Fixed in commit d940b38. Removed the unused MEMORIAL_FILE="MEMORIAL.md" assignment from line 9 — the variable was never referenced anywhere in git_wrapper.sh. Dead code gone.

ScienceIsNeato:
[fixed_in_code] Fixed in d940b38 — SKIP and PRE_COMMIT_ALLOW_NO_CONFIG now only fire for hook-invoking subcommands (commit/merge/rebase/cherry-pick/revert/am) via check_hook_env_bypasses()

ScienceIsNeato:
[fixed_in_code] Fixed in d940b38 — MEMORIAL_FILE variable removed

ScienceIsNeato:
[fixed_in_code] Fixed in this branch: added 'end_of_options=false' tracking in git_wrapper.sh main(). When '--' is encountered the loop sets end_of_options=true and skips all subsequent arguments, so pathspecs like '--no-verify' after '--' are never checked against bypass patterns. Added four shell-level regression tests in TestGitWrapperEndOfOptions including test_no_verify_after_double_dash_allowed.

ScienceIsNeato:
[fixed_in_code] Fixed in cc3f5ff: added early-return guard '[[ -z "$git_command" ]] && return 0' at the top of check_hook_env_bypasses(). When no subcommand is found (git --version, bare git), the function now returns immediately before reaching the env-var checks.

ScienceIsNeato:
[fixed_in_code] Fixed in cc3f5ff: all three export-f call sites (function blocks, subcommand blocks, npx block) now emit '[[ -n "${BASH_VERSION:-}" ]] && export -f <cmd>' instead of bare 'export -f'. This is a no-op in zsh (condition is false) and preserves bash subshell propagation for bash users.

ScienceIsNeato:
[fixed_in_code] Fixed in cc3f5ff: _escape now replaces "'" with "'\''" (the standard bash end-single-quote / escaped-quote / reopen-single-quote idiom) rather than the invalid "\'". This correctly handles any single quote in CommandMap fields within printf single-quoted strings.

ScienceIsNeato:
[fixed_in_code] Fixed in 346c009: removed sibling consolidation from both _gen_subcommand_blocks and _gen_npx_block. Each CommandMap entry now gets its own case arm with its own _msg_printf call referencing the correct forbidden name. Running 'npx prettier' now shows 'npx prettier' and 'gh run view' shows 'gh run view'.

ScienceIsNeato:
[fixed_in_code] Fixed in 346c009: _strip_marker_block now tracks whether the end marker was found. If in_block is True at the end of iteration (orphaned start, no matching end), the function returns the original content unchanged rather than silently truncating the file.

ScienceIsNeato:
[fixed_in_code] Fixed in 346c009: removed the 'gh pr view --comments' entry from COMMAND_MAPPING entirely. The entry used subcommands=('pr', 'view') which cannot distinguish 'gh pr view 42' from 'gh pr view --comments' — so it over-broadly blocked all 'gh pr view' calls. The 'gh pr checks' suggestion now mentions 'sm buff inspect' for comment browsing.

ScienceIsNeato:
[fixed_in_code] Fixed in f32e388: added `if plain:` guard before emitting the `else` branch in `_gen_function_blocks`, so the `else` block is only emitted when there are plain (non-flagged) entries. An empty `else` body is never generated, eliminating the bash 3.2 syntax error.

ScienceIsNeato:
[fixed_in_code] Fixed in f32e388: both `_ALIASES_DEST.unlink()` and `_WRAPPER_DEST.unlink()` are now wrapped in `try/except OSError` with warning messages on failure. `_mutinize_uninstall` now tracks an `errors` counter (same pattern as `_mutinize_install`) and returns `1 if errors else 0` instead of unconditionally returning 0.

ScienceIsNeato:
[fixed_in_code] Fixed in e6185ce. Added conditional leading \n to rc_block: we now prepend \n only when the rc file does not already end with \n, so GANG_MARKER is never concatenated onto the preceding line. _strip_marker_block therefore strips cleanly without eating the preceding line on uninstall. The round-trip byte-identical invariant is preserved.

ScienceIsNeato:
[fixed_in_code] Fixed in e6185ce. Emergency removal instructions now loop over all four rc candidates (~/.zshrc ~/.zprofile ~/.bashrc ~/.bash_profile) instead of hardcoding ~/.zshrc. Both _show_confirm_warning, _generate_aliases_sh header, and _gang_list output were updated.

ScienceIsNeato:
[no_longer_applicable] Barnacle-only intercept behavior was intentional — introduced in commit 77c40d1 ('make mutinize intercepts barnacle-only — no redirect hints, no mapping table'). The decision was to stop silently redirecting (which confused agents) and instead always block with a barnacle message. COMMAND_MAPPING still drives which commands are intercepted and how (function vs subcommand vs npx); it is no longer the source of redirect output, which is now uniform.

ScienceIsNeato:
[fixed_in_code] Fixed in e6185ce. Added _MUTINIZE_MARKER / _MUTINIZE_END_MARKER migration constants. _gang_uninstall now strips old sm-mutinize blocks in addition to the new sm-gang and legacy deep-hooks blocks. _gang_install detects old sm-mutinize blocks and migrates them (strips old block, then appends new sm-gang block) instead of silently installing a second block.

ScienceIsNeato:
[fixed_in_code] Fixed in e6185ce. Added errors += 1 in the OSError handler for _ALIASES_DEST chmod failure, matching the identical pattern already present for _WRAPPER_DEST. _gang_install now returns 1 when aliases.sh permissions cannot be set.

ScienceIsNeato:
[fixed_in_code] Fixed: _pip_audit_resolve_req_files now distinguishes between 'no Python project' and 'pyproject.toml/setup.py exists but no requirements.txt'. The latter now produces an accurate message advising to activate a virtual environment.

PR #215, opened 2026-05-22
fix: treat pip-audit timeout as warned skip instead of hard failure

ScienceIsNeato:
## Summary

- pip-audit fetches vuln data from PyPI over the network — in CI it reliably times out at the old 30s limit
- A **timeout is not a security finding**; previously it produced a hard gate failure and blocked the release
- Increase timeout to 120s and treat timeout as a non-blocking warning (skip the scan, log a message)
- Extract `_pip_audit_no_fix_result` helper to keep `_run_pip_audit` ≤ 100 lines
- Add `timed_out=False` to existing test mocks (MagicMock returns truthy for unknown attrs)
- Add `test_pip_audit_timed_out_warns_and_passes` to document the new behavior

## Root cause

The v1.3.0 and v1.3.1 release Quality Gate steps were failing because the release CI
environment (cold pip-audit cache, no VIRTUAL_ENV, scanning via `-r requirements.txt`)
consistently hit the 30s subprocess timeout — which was then misclassified as "pip-audit found issues".

## Test plan

- [ ] `TestRunPipAudit` — all 19 tests pass
- [ ] `sm swab --no-cache` — all 15 gates pass

### DIFF
--- a/slopmop/checks/security/__init__.py
+++ b/slopmop/checks/security/__init__.py
@@ -1276,6 +1276,12 @@ class SecurityCheck(SecurityLocalCheck):
                 True,
                 "pip-audit timed out fetching vulnerability data — skipped",
                 warned=True,
+                sarif_findings=[
+                    Finding(
+                        message="pip-audit timed out; vulnerability scan skipped",
+                        level=FindingLevel.WARNING,
+                    )
+                ],
             )
 
         try:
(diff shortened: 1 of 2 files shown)


ScienceIsNeato:
[fixed_in_code] Restored the with/assert block in test_pip_audit_json_error_success and removed the duplicate patch block from the timeout test. Fixed in commit 1f0c3d5.

ScienceIsNeato:
[fixed_in_code] Same fix as above — restored full test body. Commit 1f0c3d5.

ScienceIsNeato:
[fixed_in_code] Added explicit WARNING Finding to the timeout SecuritySubResult. Commit 1f0c3d5.

PR #221, opened 2026-05-25
Gang press discharge

ScienceIsNeato:


ScienceIsNeato:
[no_longer_applicable] Fixed in commit 92d63f2: _format_available_fields() now only adds blank line separator when gate_fields exist (line 138: if gate_fields:). Output no longer starts with blank line.

ScienceIsNeato:
[no_longer_applicable] Typo already fixed in .willville.json line 21: 'dogfooded' (correct spelling) was already in place.

ScienceIsNeato:
[no_longer_applicable] Test added in b72f3eb: test_windows_platform_returns_true_without_ps_check() patches sys.platform='win32' and verifies early return without subprocess call.

ScienceIsNeato:
[no_longer_applicable] Test added: test_enable_gate_allows_pr_context_gates() exercises the PR-context allowance logic with mocked check (is_applicable=False, skip_reason with 'PR context detected'). Tests both enable and disable transitions.

ScienceIsNeato:
[out_of_scope_ticketed] This refactoring is valid but beyond PR #221 scope. The gate enablement logic in _enable_gate() currently uses substring matching on skip_reason. A better approach would use a structured signal (enum/flag on check). Created issue to track: structured gate applicability signals for future refactoring.

ScienceIsNeato:
[fixed_in_code] Fixed in commit ad7b032: _try_suggest_config_command now returns bool as documented. When True is returned (suggestion printed), caller exits; when False, falls back to original_error. Function contract now matches implementation.

ScienceIsNeato:
[fixed_in_code] Fixed in bee180d — changed search string from 'not on a PR' to 'not on a pr' so the lowercase comparison against reason.lower() works correctly.

PR #223, opened 2026-05-25
fix: starlette vuln pin + dead-code.js knip 6 compat + ignore config fields

ScienceIsNeato:
## Summary

- Pin `starlette>=1.0.1` in the `security` extra to resolve PYSEC-2026-161 (path traversal via `StaticFiles`)
- Fix `laziness:dead-code.js` knip 6.x JSON format regression: knip 5+/6+ wraps output as `{"issues": [...]}` — the gate was treating all findings as an error because the bare-list assumption failed, producing `💥 error` instead of structured `❌ failed` findings
- Add `ignore_patterns` and `ignore_dependencies` config fields so tooling entry points (`.detoxrc.js`, `.maestro/**`, jest configs) and config-consumed deps can be suppressed without requiring a full `knip.json` — this was the root cause of an agent doom-loop in a real fogofdog session
- Extract `_parse_symbol_findings` and `_parse_member_findings` helpers to stay under the 100-line function limit

## Test plan

- [ ] `sm scour --no-cache` passes (18 passed, 1 warned pre-existing, 0 failed)
- [ ] `pip-audit` against venv no longer flags PYSEC-2026-161
- [ ] `laziness:dead-code.js` in fogofdog-frontend now shows `❌ failed` with structured findings (not `💥 error`)
- [ ] New unit tests cover knip 6 envelope format, `ignore_patterns`/`ignore_dependencies` temp config generation, and config precedence

### DIFF
--- a/.willville.json
+++ b/.willville.json
@@ -25,35 +25,20 @@
     "updated": "2026-05-25"
   },
   "agent": {
-    "status": "PR #221: coverage test added for not-on-a-PR path, awaiting CI",
-    "direction": "Fix + test the case-sensitive gate check; unblock codecov/patch",
-    "difficulties": "None",
-    "needs_human": "None",
-    "last_update": "2026-05-25T09:28:00Z",
+    "status": "PR #223 open: starlette vuln pin + dead-code.js knip 6 compat",
+    "direction": "Two fixes: PYSEC-2026-161 starlette pin and dead-code.js knip 6 JSON format compat + ignore_patterns/ignore_dependencies config fields to stop agent doom-loops on tooling file false positives",
+    "last_update": "2026-05-25T00:00:00Z",
     "actions": [
-      { "name": "Fixed case-sensitive PR context check (config.py:380)", "status": "done" },
-      { "name": "Resolved Bugbot thread PRRT_kwDORBxXu86EgBGI", "status": "done" },
-      { "name": "Added coverage test for not-on-a-PR branch path", "status": "done" },
-      { "name": "Wait for CI to confirm codecov/patch green", "status": "in_progress" },
-      { "name": "Merge PR #221", "status": "planned" }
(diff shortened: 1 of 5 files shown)


ScienceIsNeato:
[fixed_in_code] Fixed in commit 834393b. When ignore_patterns or ignore_dependencies are set without a knip_config, the temp config now checks for knip.json, .knip.json, knip.ts, and .knip.ts in the project root and adds an `extends` key if found. This preserves the repo's full knip config (entry points, plugins, project settings) while still applying the sm-supplied ignore overrides.

ScienceIsNeato:
[invalid_with_explanation] The dead-code.js changes are present in the PR diff. This branch contains three commits on top of main: b388fa5 (starlette pin), 4b74bb5 (dead-code.js knip 6 compat + ignore_patterns/ignore_dependencies + helper extraction), and 834393b (extends fix from your sibling thread). The PR description accurately describes all changes. The diff viewer may have shown only the most recent commit — the full branch diff includes the dead-code.js work.

ScienceIsNeato:
[fixed_in_code] Fixed in commit b134ecf. knip does not define extends in its config schema — using it was wrong. The approach is now: read the existing JSON knip config (knip.json, knip.jsonc, .knip.json, .knip.jsonc), merge ignore_patterns and ignore_dependencies into it, and write the merged result as _sm_knip.json. This preserves all existing entry points, plugins, and project settings. TS/JS configs are skipped as they are not safely parseable; users with those configs should set knip_config explicitly.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 72305a4: _load_repo_knip_config now reads the file as text and strips // line comments via re.sub before calling json.loads, so typical JSONC files (e.g. knip.jsonc with // entry points comments) parse correctly and their settings are preserved in the merge. Added test_temp_config_merges_knip_jsonc_with_comments to cover this path.

PR #224, opened 2026-05-26
chore: update willville after PR #223 merge

ScienceIsNeato:
Updates .willville.json to reflect PR #223 merged and starlette blocker cleared.

### DIFF
--- a/.willville.json
+++ b/.willville.json
@@ -23,16 +23,9 @@
     "updated": "2026-05-26"
   },
   "agent": {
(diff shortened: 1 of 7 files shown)


ScienceIsNeato:
[fixed_in_code] Fixed in slopmop/core/lock.py: _pid_alive now checks for ERROR_ACCESS_DENIED (5) on Windows OpenProcess failures. When access is denied, the process is treated as alive (similar to POSIX PermissionError), preventing false-positive stale lock detection on Windows for processes running as different users or elevated/protected processes.

PR #227, opened 2026-05-27
feat: sailing mode + success-path guidance for agents

ScienceIsNeato:
- `sm sail` now activates SAILING mode on entry and persists it across
  calls until PR_READY, so agents can drive all the way to a green PR
  without stopping to ask what to do next.

- Two modes: ITERATING (default — surface results, await human) vs
  SAILING (human approved — emit exact git/gh commands at every step).

- Success paths now emit `next:` guidance, filling the gap where swab
  or scour passes used to go silent and leave agents guessing.

- SCOUR_CLEAN handler emits exact command strings
  (`git push -u origin HEAD && gh pr create --fill`) rather than
  vague descriptions.

- PR_READY resets to ITERATING and tells the agent to surface the PR
  to the human rather than suggesting `sm buff finalize --push`.

- `SailMode` enum + `read_sail_mode`/`write_sail_mode` in state_store.
- `RunReport.next_step` computed from level + workflow_state + sail_mode.
- Console and porcelain adapters emit next_step on every clean pass.

### DIFF
--- a/.cursor-plugin/skills/sm-sail.md
+++ b/.cursor-plugin/skills/sm-sail.md
@@ -33,9 +33,9 @@ sm sail → runs swab
   buff all-green           → ⛵ PR ready for human review
 ```
 
-## Note on iterating-mode guidance
+## Note on tacking-mode guidance
 
-When `sm swab` passes outside of sail (iterating mode), the output includes
+When `sm swab` passes outside of sail (tacking mode), the output includes
 "share results with the human, await the next instruction." That guidance comes
 from `sm swab` directly — `sm sail` is always in SAILING mode.
 
--- a/.cursor-plugin/skills/sm-sail.md
+++ b/.cursor-plugin/skills/sm-sail.md
@@ -1,16 +1,16 @@
 ---
 name: sm-sail
 description: >-
-  Auto-advance the slop-mop workflow. Use when unsure what to run next,
-  or to let slop-mop drive the iteration loop automatically. Triggers
-  when the user says "sail", "what should I run?", "advance the workflow",
-  "next step", or "keep going" in a project using slop-mop.
+  Drive the slop-mop workflow toward a green, buffed PR. Use when the human
+  says "sail", "ship it", "take it home", or approves the work and wants it
+  shipped. Also use when unsure what to run next in a slop-mop project.
 ---
 
-# sm sail — workflow auto-advance
+# sm sail — workflow autopilot
 
-`sm sail` reads the current workflow state and dispatches the right verb
-automatically. Use it when you don't know whether to swab, scour, or buff.
+`sm sail` reads the current workflow state and mode, runs the next step or
+emits the exact command to run, then exits. Call it again after following
+its instruction.
 
 ## Prerequisite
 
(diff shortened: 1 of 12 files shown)


ScienceIsNeato:
[fixed_in_code] Fixed in eb53100. Rewrote _compute_next_step to check level first ('swab' or 'scour') before any workflow_state fallback. A direct 'sm scour' run has workflow_state=IDLE at report-build time (pre-hook), so the old 'or workflow_state in (...)' guard made scour passes emit swab guidance. Now level is the sole signal; workflow_state and the WorkflowState import are removed from the function entirely.

ScienceIsNeato:
[fixed_in_code] Fixed in eb53100. Removed the ITERATING branch from _sail_swab_clean entirely. Since cmd_sail always writes SailMode.SAILING on entry, read_sail_mode in the handler always returned SAILING — the else branch was unreachable in production. Simplified to a single unconditional commit instruction. Also removed the stale read_sail_mode import from sail.py.

ScienceIsNeato:
[fixed_in_code] Fixed in eb53100. Both issues addressed: (1) ITERATING branch removed — sail always emits a single unconditional commit instruction since cmd_sail always sets SAILING mode. (2) PR_READY reset to ITERATING is preserved — write_sail_mode(ITERATING) in _sail_pr_ready fires before the next cmd_sail call activates SAILING, so a fresh cycle starts correctly.

ScienceIsNeato:
[fixed_in_code] Fixed in eb53100. Removed test_swab_clean_uncommitted_iterating_gives_share_guidance — it was testing an impossible production state by independently mocking write_sail_mode (no-op) and read_sail_mode (ITERATING), which can't occur in real usage. Renamed the remaining sailing test to test_swab_clean_uncommitted_gives_exact_commit_command and removed the stale read_sail_mode mock.

ScienceIsNeato:
[fixed_in_code] Fixed in eb53100. Updated the Cursor skill doc to remove the 'Iterating (default)' mode table and replace it with accurate descriptions: sm sail is always SAILING; the iterating-mode commit/share guidance comes from sm swab outputs when run directly.

ScienceIsNeato:
[fixed_in_code] Fixed in eb53100. Updated the Claude template command doc with the same correction — removed 'Iterating (default)' description and clarified that sm sail activates SAILING mode on every invocation.

ScienceIsNeato:
[fixed_in_code] Fixed in eb53100. Updated commands/sm-sail.md: removed the 'Iterating (default)' mode section and replaced it with a clear note — sm sail activates SAILING mode on entry, iterating-mode guidance lives in sm swab output.

ScienceIsNeato:
[fixed_in_code] Fixed in eb53100. Removed workflow_state from _compute_next_step's signature entirely (level is now the sole signal), so the docstring/behavior mismatch is gone. The docstring now accurately reads: 'Returns None for unrecognised levels or when level is not provided.'

ScienceIsNeato:
[fixed_in_code] Fixed in 8e22c94. Made the scour branch in _compute_next_step mode-aware: SAILING mode emits 'git push ... then sm sail'; ITERATING mode emits 'share results with the human and await the next instruction.' Now mirrors swab's mode-aware behaviour exactly.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 4cddfc2. Changed PorcelainAdapter guard from `not report.actionable` to `not report.failed and not report.errored`, matching the semantics of `all_passed` (warnings don't block next_step). Added test `test_next_step_emitted_when_warnings_present` to cover this path.

PR #237, opened 2026-05-31
feat(sail): make PR-open flow watch CI before reporting

ScienceIsNeato:
## What

Follow-up flow update: after opening/updating a PR, agents should watch CI before reporting status.

## Summary

- Set PR_OPEN next action to `sm buff watch <PR#>` in the workflow state machine.
- Updated sail wording and README loop to show watch-first behavior.
- Regenerated workflow docs to keep state table/diagram aligned.
- Updated state-machine tests for the new default action.

## Validation

- `python -m pytest tests/unit/test_state_machine.py -q`
- `sm swab --no-tty --output-file .slopmop/last_swab.json`

PR #238, opened 2026-05-31
feat: unify all verbs under the v3 machine-interface envelope

ScienceIsNeato:
## Summary
- Every JSON-emitting verb now wraps its output in one versioned envelope (`slopmop/v3`), replacing the two divergent conventions (validation's flat shape + status's ad-hoc dict). An agent can know the exact shape of every response before running anything.
- Adds discovery without execution — `sm schema [verb]`, `sm capabilities`, and a universal `<verb> --describe` — plus a published per-verb data schema and a CI conformance test that validates each verb's real output against its composed schema (passing and failing runs).
- Removes the dead `slopmop/v2` marker and the unused `RunReport.schema_version`; documents the contract in `DOCS/MACHINE_INTERFACE.md`.

## Note on the diff
Branched off `feat/captain-verb` (open PR #237), so until #237 merges this PR's diff includes #237's commits. They drop out once #237 lands.

## Test plan
- [x] `pytest tests/unit` — 3011 passed
- [x] `sm scour --no-auto-fix` — NO SLOP DETECTED (18 passed, 2 non-blocking warnings)
- [ ] CI green

### DIFF
--- a/slopmop/cli/capabilities.py
+++ b/slopmop/cli/capabilities.py
@@ -31,6 +31,13 @@ from slopmop.reporting.envelope import (
 # the parser registration order. ``data_schema`` is resolved at runtime
 # from the packaged schemas so this list never claims a contract that
 # isn't shipped — see _verb_entries().
+#
+# Drift guards (tests/unit/test_capabilities.py): the verb *set* is
+# pinned to the parser (test_catalog_lists_every_registered_verb),
+# ``group`` to an enum, and ``formats`` to each verb's actual argparse
+# output flags (test_catalog_formats_track_argparse_output_flags).
+# ``summary`` and ``exit_codes`` are editorial with no second source, so
+# this literal is their truth; exit_codes shape is structurally checked.
 _VERB_CATALOG: List[Dict[str, object]] = [
     {
         "name": "swab",
--- a/slopmop/cli/capabilities.py
+++ b/slopmop/cli/capabilities.py
@@ -103,8 +103,9 @@ _VERB_CATALOG: List[Dict[str, object]] = [
         "group": "escalation",
         "formats": ["human", "json"],
         "exit_codes": {
-            "0": "summons accepted",
+            "1": "summons answered — orders received, loop paused",
             "2": "justification insufficient — standing order read back",
+            "3": "no human at an interactive terminal — re-run where one can answer",
         },
     },
     {
--- a/slopmop/cli/capabilities.py
+++ b/slopmop/cli/capabilities.py
@@ -1,0 +1,317 @@
+"""`sm capabilities` — the discovery catalog.
+
+One read teaches an agent the entire surface: the installed slop-mop
+version, every verb with its output contract (group, accepted formats,
+exit-code meanings, and the ``$id`` of its packaged data schema when it
+has one), and every registered gate with its metadata and applicability
+to *this* project.
+
+Like ``sm schema``, this is self-description without execution — it runs
+no gates. The payload itself rides the standard envelope, so an agent
+that has read ``sm schema capabilities`` already knows this command's
+exact shape before running it.
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+from pathlib import Path
+from typing import Dict, List, Optional
+
+from slopmop import __version__
+from slopmop.reporting.envelope import (
+    ENVELOPE_SCHEMA_VERSION,
+    Status,
+    available_data_schemas,
+    build_envelope,
+)
+
+# Curated verb catalog. Order mirrors the workflow an agent walks, not
+# the parser registration order. ``data_schema`` is resolved at runtime
+# from the packaged schemas so this list never claims a contract that
+# isn't shipped — see _verb_entries().
+_VERB_CATALOG: List[Dict[str, object]] = [
+    {
+        "name": "swab",
+        "summary": "Quick validation — runs every swab-level gate (every commit).",
+        "group": "core",
+        "formats": ["human", "json", "porcelain", "sarif"],
+        "exit_codes": {
+            "0": "all gates passed",
+            "1": "one or more gates failed",
+            "2": "could not run (bad args, missing dependency)",
+        },
+    },
+    {
+        "name": "scour",
+        "summary": "Thorough validation — runs all gates (PR readiness, superset of swab).",
+        "group": "core",
+        "formats": ["human", "json", "porcelain", "sarif"],
+        "exit_codes": {
+            "0": "all gates passed",
+            "1": "one or more gates failed",
+            "2": "could not run (bad args, missing dependency)",
+        },
+    },
+    {
+        "name": "sail",
+        "summary": "Auto-advance the workflow — read state and do the next obvious step.",
+        "group": "workflow",
+        "formats": ["human", "json"],
+        "exit_codes": {
+            "0": "step completed",
+            "1": "step ran but found blocking problems",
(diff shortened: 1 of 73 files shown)


ScienceIsNeato:
[fixed_in_code] Fixed in df55173. _failed_results_from_scour_artifact (slopmop/cli/refit.py) and _summarise_failure_artifact (slopmop/cli/_refit_iteration.py) now unwrap the v3 envelope's `data` key before reading `results`, so a failing scour produces a populated remediation plan again. Regression coverage: tests/unit/test_refit.py::test_build_plan_orders_failed_gates_and_derives_commit_messages and tests/unit/test_refit_iterate.py now feed enveloped artifacts.

ScienceIsNeato:
[fixed_in_code] Fixed in df55173. sm buff inspect (slopmop/cli/buff.py) and sm refit (_emit_protocol in slopmop/cli/refit.py) now write the same v3 envelope to the --output/--json-file mirror that they print to stdout, so a pipeline capturing the file sees the documented contract. refit's persisted protocol file stays bare since it's internal state, not the user-facing mirror. Regression coverage: tests/unit/test_buff_inspect_and_status.py::test_cmd_buff_json_mode asserts file == stdout envelope; test_refit.py and test_refit_continue.py assert the mirror is enveloped (data == protocol).

ScienceIsNeato:
[fixed_in_code] Fixed in 5fc2a9a: _read_actionable_results now unwraps the v3 envelope's data slot before reading results (data_raw = json_doc.get('data'); raw_results = data.get('results')). Added regression test test_read_actionable_results_unwraps_v3_envelope.

ScienceIsNeato:
[fixed_in_code] Fixed in 5fc2a9a: implemented JSON for buff status/watch. cmd_buff_status now accepts json_output; status emits the v3 envelope (command=buff) and watch streams human progress then prints the final envelope. The capabilities catalog claim of json for buff is now accurate. Added regression tests test_status_json_clean_emits_envelope and test_status_json_feedback_blocked_emits_fail_envelope.

ScienceIsNeato:
[fixed_in_code] Fixed in 02a329e: the capabilities catalog for wake-angry-drunk-captain now documents the real exit codes — 1 (summons answered, orders received, loop paused), 2 (justification insufficient), 3 (no human at an interactive terminal). The previous 0/2 entries were wrong; the verb never returns 0 (see EXIT_SUMMONED=1/EXIT_REFUSED=2/EXIT_NO_CAPTAIN=3 in slopmop/cli/captain.py).

ScienceIsNeato:
[fixed_in_code] Fixed in 02a329e: slopmop/schemas/v3/data/buff.json is now a oneOf of two $defs — 'triage' (the inspect/scanning payload: source, summary, actionable, hard_failures, next_steps, pr_resolution, pr_feedback) and 'status' (overall_state plus checks/feedback). This mirrors the established oneOf convention in doctor.json and wake-angry-drunk-captain.json. The two branches are mutually exclusive (status requires overall_state, triage never has it), and a new conformance test test_buff_status_output_conforms_to_schema validates the status envelope against the composed schema.

ScienceIsNeato:
[fixed_in_code] Fixed in 02a329e: refit's _emit_protocol now sets envelope status to Status.INFO only when exit_code == 0, and Status.FAIL otherwise. A blocked outcome (missing plan, wrong phase, etc.) returns exit_code 1 and now reports status 'fail' instead of 'info', so parsers treating info as non-blocking no longer miss the failure. Updated test_refit_output_conforms_to_schema to assert status 'fail' for the missing_plan exit-1 case.

ScienceIsNeato:
[fixed_in_code] Fixed in 3f1440f. The pre-push merged-branch guard now reads the refs Git feeds on stdin (<local ref> <local sha> <remote ref> <remote sha>) and inspects each pushed branch derived from refs/heads/*, rather than only querying HEAD. A push like 'git push origin merged-feature:merged-feature' from another checkout is now guarded; deletions (all-zero local sha) are skipped. Updated test_generate_pre_push_hook_script to assert the stdin-reading loop and that git symbolic-ref is no longer used.

slopmop/cli/capabilities.py
```diff
@@ -0,0 +1,318 @@
...
+    available_data_schemas,
+    build_envelope,
+)
+
+# Curated verb catalog. Order mirrors the workflow an agent walks, not
+# the parser registration order. ``data_schema`` is resolved at runtime
+# from the packaged schemas so this list never claims a contract that
+# isn't shipped — see _verb_entries().
+_VERB_CATALOG: List[Dict[str, object]] = [
+    {
```
ScienceIsNeato, line 42: should it worry me that this is a long literal string instead of generated? If this is the ultimate source of truth, I'm not too worried. Otherwise, I'm very worried

ScienceIsNeato:
[needs_human_feedback] Good instinct to check. The short answer: for the fields you'd worry about duplicating, this literal is *not* the source of truth — they're conformance-tested or runtime-resolved, so they can't drift into a lie. For the rest, it is the source of truth by design, which puts you in your "not too worried" case.

Breakdown by field:

- `name` — the parser is the truth, not this list. `test_catalog_lists_every_registered_verb` diffs the catalog against the parser's registered subcommands and fails CI the moment a verb is added/removed without updating here. So the verb *set* can't silently drift.
- `data_schema` — not present in this literal at all. It's computed at runtime in `_verb_data_schema_id()` from `available_data_schemas()` (what actually ships under schemas/v3/data/), and `test_verb_data_schema_refs_match_shipped_schemas` guards it. The catalog physically cannot advertise a schema that isn't shipped.
- `group` — editorial, but pinned to an enum by `test_catalog_groups_are_in_the_known_vocabulary`.
- `summary`, `formats`, `exit_codes` — genuinely editorial. This literal *is* the ultimate source of truth; nothing else documents per-verb exit-code semantics or accepted formats in machine-readable form.

The one honest soft spot: `formats` and `exit_codes` describe real runtime behavior (argparse flags, actual return codes) but have no second source to diff against, so they can fall out of date if a verb changes its flags or exit codes. `formats` is partially derivable from each verb's argparse; `exit_codes` isn't derivable at all, so the literal is canonical there by necessity.

Leaving this thread open — if you want a drift-guard for `formats`/`exit_codes` I can add one (introspect argparse for formats; exit_codes would need a lightweight per-verb registry). Otherwise it's safe to resolve.

ScienceIsNeato:
[fixed_in_code] Closed the soft spot in 3fd482a. The catalog is now drift-guarded field by field:

- `name` → conformance-tested against the parser (existing).
- `data_schema` → runtime-resolved from shipped schemas (existing).
- `group` → enum-pinned (existing).
- `formats` → NEW: test_catalog_formats_track_argparse_output_flags cross-checks each verb's declared formats against its actual argparse output dests (json_output/sarif_output/porcelain). Keyed by dest, not option string, so `config --json` (an input flag) is correctly excluded. sarif/porcelain are opt-in only → biconditional; json is a forward implication because json-only verbs (schema, capabilities) emit it with no flag.
- `exit_codes` → NEW: structural guard (numeric keys in range, non-empty meanings). No second source exists to diff against, so this literal stays canonical there by necessity — but garbage/typos now fail CI.

So `summary` and `exit_codes`-semantics remain editorial-by-design (the "ultimate source of truth" case you're fine with), and everything that mirrors another source now fails CI on drift. Verified the new guards bite by planting drift (claiming sarif on a verb with no --sarif flag, and an "o"-for-"0" typo) — both caught.

PR #247, opened 2026-06-02
fix(type-completeness): honor project pyright rule overrides (#245)

### LINKED ISSUE #245: type-completeness gate: no easy way to exclude reportUnknown* noise originating in untyped dependencies (exclude_paths doesn't reach consumer call sites)
## Summary

The `overconfidence:type-blindness.py` gate (strict pyright type-**completeness**) gives a project **no easy way to exclude type-completeness noise that originates in dependencies it doesn't own** — untyped third-party packages and git submodules. The result is hundreds of `reportUnknown*` errors a project cannot fix without annotating code in other repos, and the existing `exclude_paths` knob does not suppress them.

This is distinct from a project's own missing annotations (which the mypy `missing-annotations` gate covers and which the project *can* fix). Here the unknowns come from outside.

## Why `exclude_paths` doesn't help

`[tool.slopmop] exclude_paths` excludes those paths from being **scanned**. But type-completeness errors are reported at the **consumer's call site**, not in the excluded dependency. Example from a real repo:

```toml
[tool.slopmop]
exclude_paths = ["ganglia-common", "ganglia-studio"]   # submodules, not scanned
```

```
src/ganglia_core/server.py
  L194: [reportUnknownMemberType] Type of "print_error" is partially unknown
  L67:  [reportUnknownMemberType] Type of "add_middleware" is partially unknown
src/ganglia_core/dictation/stt_provider.py
  L236: [reportUnknownMemberType] Type of "recognize_google" is unknown
```

`print_error` comes from the excluded `ganglia_common` submodule; `add_middleware` from `fastapi`; `recognize_google` from `speech_recognition`. All three are untyped upstream — but the error lands on **our** file, so `exclude_paths` (which targets the *dependency's* path) never matches it. There is no per-rule or per-origin escape hatch.

Adding a project `pyrightconfig.json` with `extraPaths` so pyright can resolve the sibling packages makes **no difference** (in testing, 534 → 530 errors): pyright then *finds* the modules, but their symbols are genuinely unannotated, so completeness still can't be satisfied.

## Impact

On any repo that depends on untyped packages — common ones like `fastapi`, `speech_recognition`, `whisper`, `stripe`, or an in-house submodule without `py.typed` — the gate produces a large, **un-actionable** error count. The only ways out today are:
- annotate code in repos you don't own (often impossible), or
- disable the whole gate (loses the *real* type-error detection it also provides).

Neither is good. Teams end up disabling a useful gate wholesale.

## Requested options (any one would resolve this)

1. **First-party-only completeness**: a config flag to only enforce `reportUnknown*` completeness on symbols *defined in the project's own source*, treating unknowns whose root is an external/untyped module as allowed.
2. **A `strict: false` / "standard mode" toggle that's documented and easy to reach per-gate** — keep real type-error detection (bad imports, assignment/arg-type mismatches) while dropping the `reportUnknown*` completeness demand. (The check appears to read a `strict` flag internally; surfacing it as a first-class, documented per-gate option would be enough.)
3. **Honor pyright's own suppression** — e.g. respect `useLibraryCodeForTypes` / a per-path `reportUnknownMemberType: "none"` override in the project's `pyrightconfig.json` so untyped-dependency call sites can be scoped out without disabling the gate.

## Environment

- slopmop 2.0.1 (pipx, Python 3.14), pyright via the gate
- Repo depends on untyped submodule (`ganglia_common`) + untyped third-party (`fastapi`, `whisper`, `speech_recognition`, `stripe`)

## Related

- Downstream barnacle where this forced a gate disable: ScienceIsNeato/ganglia-core#9

ScienceIsNeato:
## Summary

The `overconfidence:type-blindness.py` gate (strict pyright completeness) gave projects **no way to suppress `reportUnknown*` noise originating in untyped dependencies** — third-party packages and submodules without `py.typed`. It generates an overlay config that `extends` the project's own pyright config, but then **unconditionally forced all five completeness rules to `error`**, clobbering any suppression the project set. And `exclude_paths` can't help: the error lands on the *consumer's* call site, not the excluded dependency. Teams were disabling a useful gate wholesale.

## Fix

The gate now honors the project's own pyright config:

- **Auto-detects** a project-owned `pyrightconfig.json` (not only an explicitly configured one) and `extends` it, so its settings actually apply.
- **In strict mode, only injects a completeness rule the project hasn't already configured** — checking both top-level keys and per-path `executionEnvironments`. So a repo can set `reportUnknownMemberType: "none"` globally or scoped to untyped call sites, and the gate respects it, while real type errors and any unmentioned completeness rules stay enforced.

This is the "honor pyright's own suppression" option from the report. An explicitly configured base config is still trusted as-is (no scope change); auto-detected configs keep source-dir include scoping so the gate doesn't balloon to the whole tree.

Closes #245

## Test plan

- [x] `sm scour` clean (18 passed, 1 pre-existing non-blocking config-debt warning)
- [x] New tests: top-level rule override honored; per-path `executionEnvironments` override honored; all rules still forced when the project has no pyright config
- [x] Existing overlay/extends tests still pass (explicit-config behavior unchanged)
- [ ] CI green

### DIFF
--- a/.willville.json
+++ b/.willville.json
@@ -23,8 +23,8 @@
     "updated": "2026-05-26"
   },
   "agent": {
-    "status": "Fixed the secret scanner choking on repos with giant vendored and dependency folders — it was crawling everything and timing out. Now it skips the heavy stuff and finishes in a blink",
-    "direction": "Folding this into the open review with the rest of the friction fixes, then getting it all into a release",
+    "status": "Type checker was burying people under errors coming from other people's untyped libraries, with no way to quiet them short of switching the whole check off. Now it respects a project's own settings so they can hush just the noisy bits",
+    "direction": "Getting this reviewed and shipped alongside the other friction fixes",
     "last_update": "2026-06-01T00:00:00Z"
   },
   "queue": {
--- a/slopmop/checks/python/type_checking.py
+++ b/slopmop/checks/python/type_checking.py
@@ -141,6 +141,20 @@ def _detect_venv_path(project_root: str) -> Tuple[Optional[str], Optional[str]]:
     return None, None
 
 
+def _strip_jsonc(text: str) -> str:
+    """Make a JSONC document parseable by ``json.loads``.
+
+    pyright config files are JSONC: they may contain ``//`` line comments,
+    ``/* */`` block comments, and trailing commas — all of which ``json.loads``
+    rejects. Strip them so a project's pyrightconfig.json is honored instead of
+    silently failing to parse (which would discard its rule overrides).
+    """
+    text = re.sub(r"/\*.*?\*/", "", text, flags=re.DOTALL)  # block comments
+    text = re.sub(r"//[^\n]*", "", text)  # line comments
+    text = re.sub(r",(\s*[}\]])", r"\1", text)  # trailing commas
+    return text
+
+
 def _detect_source_dirs(project_root: str) -> List[str]:
     """Detect which directories contain Python source code."""
     candidates = ["src", "slopmop", "lib", "app"]
@@ -300,8 +314,7 @@ class PythonTypeCheckingCheck(BaseCheck, PythonCheckMixin):
             return {}
         try:
             text = path.read_text(encoding="utf-8")
-            stripped = re.sub(r"//[^\n]*", "", text)
-            data: Any = json.loads(stripped)
+            data: Any = json.loads(_strip_jsonc(text))
         except (json.JSONDecodeError, OSError):
             return {}
         return cast(Dict[str, Any], data) if isinstance(data, dict) else {}
--- a/slopmop/checks/python/type_checking.py
+++ b/slopmop/checks/python/type_checking.py
@@ -24,7 +24,6 @@ would have to guess. This gate eliminates that guessing.
 
 import json
 import os
-import re
 import shutil
 import time
 from collections import Counter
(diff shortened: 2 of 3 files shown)


ScienceIsNeato:
[fixed_in_code] Fixed in 67c3360. Added _strip_jsonc() which removes /* */ block comments and trailing commas in addition to // line comments before json.loads, so a JSONC pyrightconfig.json now parses instead of silently returning {} (which would have emptied owned_rules and re-forced every rule to error). Added a regression test (test_strict_honors_override_in_jsonc_config) using a config with a block comment, a line comment, and a trailing comma.

ScienceIsNeato:
[fixed_in_code] Fixed in 17225fd. Replaced the regex stripper with a character-level scanner that tracks double-quoted (escape-aware) JSON strings and only strips // and /* */ comments + trailing commas in structural position — never inside string values. Verified: a config with include ['src/**'], an executionEnvironments root 'packages/*/dist' (the /* + */ that broke the regex), and a 'http://example' value now parses correctly and the rule override is honored. Regression test test_strict_honors_override_in_jsonc_config now includes these glob patterns and a // inside a string.

ScienceIsNeato:
[fixed_in_code] Fixed in 6cef96d. _completeness_rules_owned_by_project now counts only top-level rule settings as global ownership; executionEnvironments entries are no longer treated as ownership. So when a project scopes reportUnknown* to 'none' for one path, the gate still forces top-level 'error' and pyright applies the per-path 'none' on top of it for just that root — enforcement holds everywhere else. Updated the test (test_per_path_override_keeps_global_enforcement) to assert the rule stays 'error' at top level given a per-path-only override.

PR #250, opened 2026-06-03
refactor(gates): move network/expensive gates from swab to scour

ScienceIsNeato:
## Summary

The `SWAB` contract is documented as **"fast, local, no network or PR context."** Three gates violated it and are moved to `SCOUR` — which the agent rails (`_shared/core.md`) *already* route their tools to (`bandit`/`detect-secrets` → `sm scour`, `jscpd`/duplication → `sm scour`). This was code drifting from the project's own documented contract.

| Gate | Why it leaves swab |
|---|---|
| `myopia:vulnerability-blindness.py` | **Network**: semgrep `--config=auto` fetches rulesets from semgrep.dev on every run. Its docstring already said "Level: scour" but the `level` ClassVar was never set, so it ran on swab against its own stated intent. Security scanning is a pre-push concern; its sibling `dependency-risk.py` is already scour. |
| `laziness:repeated-code` | **Expense**: jscpd clone detection (~4–16s), slow-changing signal. |
| `myopia:string-duplication.py` | **Expense**: cross-file token scan (~10–50s). |

`untested-code` (the full test suite, the other big swab cost) is **deliberately left on swab** — "do the tests pass" is the most foundational per-commit signal; the right fix there is caching/impact-selection, not releveling.

Result: swab drops from 16 → 13 gates; scour is unchanged (superset). Teams wanting any of these per-commit can opt in via `run_on` in config.

## Test plan

- [x] `swab` (via checkout) now runs 13 gates — the three are excluded
- [x] `scour` runs 18 and stays green — the three still run there
- [x] Registry probe confirms all three resolve to SCOUR by default (no config override)
- [x] Unit test updated: swab-off example now uses an actual swab gate
- [x] Docker integration security assertions updated to verify the gates are scour-only
- [ ] CI green

### DIFF
--- a/.willville.json
+++ b/.willville.json
@@ -23,8 +23,8 @@
     "updated": "2026-05-26"
   },
   "agent": {
-    "status": "Tidying up two things at once: one shared way to read config files with comments (we had three slightly-broken copies), and making sure friction reports always get tagged so they don't get lost",
-    "direction": "Wrapping these up; one unrelated security alert on a borrowed library is in the way and needs a call on how to handle it",
+    "status": "Moved the slow and network-using checks out of the quick every-commit pass and into the thorough pre-PR pass, so the quick loop stays fast and stays offline like it's supposed to",
+    "direction": "Getting this reviewed and shipped; also flagging that my local copy of the tool is out of date",
     "last_update": "2026-06-01T00:00:00Z"
   },
   "queue": {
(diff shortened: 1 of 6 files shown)


ScienceIsNeato:
[fixed_in_code] Fixed in this push. test_key_gates_run no longer expects vulnerability-blindness.py in the all-pass swab output (it's scour-only now); swapped it for dead-code.py, a genuine swab gate. The other docker security assertions (all-fail, mixed) were already updated to verify the gate is scour-only and does not run in swab.

PR #277, opened 2026-06-12
fix: pin actions to SHAs, enable PyPI attestations, fix detect-secrets baseline

### LINKED ISSUE #253: [barnacle] detect-secrets .secrets.baseline can't allowlist false positives (config replay drops default filters)
### Barnacle Summary
detect-secrets .secrets.baseline can't allowlist false positives (config replay drops default filters)

### Current Behavior
_run_detect_secrets writes a throwaway baseline carrying the real baseline's plugins_used/filters_used and re-scans via --baseline. That replay (a) loses detect-secrets' default heuristic filters, inflating detections (~24 -> ~92), and (b) the re-scan's (path, hashed_secret) don't match the baseline's results, so _filter_known_secrets suppresses nothing. The sanctioned baseline remediation neither filters nor stays quiet; it makes the scan noisier.

### Expected Behavior
A committed .secrets.baseline (from `detect-secrets scan`) pointed at via config_file_path should allowlist its recorded false positives so the gate passes.

### Command
```bash
sm scour -g myopia:dependency-risk.py
```

### Reproduction Steps
1. detect-secrets scan > .secrets.baseline
2. set [tool.slopmop.myopia.gates.'dependency-risk.py'] config_file_path = '.secrets.baseline'
3. sm scour -g myopia:dependency-risk.py  # detections not filtered, count inflated

### Things Tried
- Regenerated baseline plainly (plugins_used=27 -> inflation).
- Only worked after stripping plugins_used/filters_used to [] (skips the throwaway, runs a plain default scan) AND adding .secrets.baseline to exclude_dirs (its sha1 hashes flag as high-entropy strings).

### Output Excerpt
```text
tests.yml:20,28 + ~90 more flagged WITH a valid baseline present; 'No secrets' only achievable by emptying plugins_used/filters_used.
```

### Impact
- Barnacle: true
- Blocker Type: non-blocking
- Affected Workflow: scour
- Source Repository: /Users/pacey/Documents/SourceCode/ganglia_repos/ganglia-core
- Gate: myopia:dependency-risk.py

### Environment Metadata
```json
{
  "agent": {
    "name": "pacey@MacBook-Pro.local",
    "source": "unknown"
  },
  "cwd": "/Users/pacey/Documents/SourceCode/ganglia_repos/ganglia-core",
  "filed_at": "2026-06-05T02:52:24.897940+00:00",
  "platform": "macOS-26.3.1-arm64-arm-64bit-Mach-O",
  "python_version": "3.14.3",
  "repo": {
    "branch": "main",
    "commit": "5af0352655b5f9549df6d7afb0190909dbd8a486",
    "dirty": true,
    "remote": "git@github.com:ScienceIsNeato/ganglia-core.git",
    "root": "/Users/pacey/Documents/SourceCode/ganglia_repos/ganglia-core"
  },
  "schema": "slopmop/barnacle-issue/v1",
  "slopmop_version": "2.0.2"
}
```

### LINKED ISSUE #269: Release: emit + verify PEP 740 attestations on PyPI (Trusted Publishing already wired)
## Finding
Trusted Publishing is already configured in `release.yml` (`publish-pypi` job has `environment: pypi`, `permissions: id-token: write`, and `pypa/gh-action-pypi-publish@release/v1`). But the published **2.3.1** artifacts carry **no PEP 740 attestation**:

| Artifact | provenance |
|---|---|
| `slopmop-2.3.1-py3-none-any.whl` | none |
| `slopmop-2.3.1.tar.gz` | none |

(Confirmed via `https://pypi.org/pypi/slopmop/2.3.1/json` — both `urls[].provenance` are `null`.)

So consumers can't cryptographically verify the PyPI artifact was built from the release commit in GitHub Actions — the one real gap in an otherwise airtight release contract (tag = PyPI = `_version.py` = 2.3.1, test-enforced).

## Current publish step
```yaml
- name: Publish to PyPI
  uses: pypa/gh-action-pypi-publish@release/v1
  with:
    skip-existing: true
```

## Proposed change
1. Pin the action to a recent **SHA** (supply-chain hygiene + determinism; `release/v1` is a moving tag) on a version that supports attestations-by-default with Trusted Publishing.
2. Set `attestations: true` explicitly so it never silently depends on the resolved default:
   ```yaml
   - name: Publish to PyPI
     uses: pypa/gh-action-pypi-publish@<pinned-sha>  # >= v1.11
     with:
       skip-existing: true
       attestations: true
   ```
3. Add a **post-publish verification** step (and a line in `DOCS/RELEASING.md` Post-Publish Checklist): assert `urls[].provenance != null` on the PyPI JSON for the new version, failing the release if attestations are missing.

## Why now
slop-mop is positioning as a trustworthy CI gatekeeper; signed provenance is exactly the property that audience verifies. Low effort, high trust payoff. Only validatable on a real release, hence tracking here rather than a speculative PR.

## Acceptance
- Next release's wheel + sdist show a non-null `provenance` on PyPI.
- Release fails loudly if attestations are absent.

### LINKED ISSUE #271: CI hardening: pin GitHub Actions to commit SHAs in workflows
zizmor / CodeRabbit flag that `.github/workflows/slopmop-sarif.yml` (and likely the other workflows) pin actions to mutable `@v*` tags rather than commit SHAs:

- `actions/checkout@v5`
- `actions/setup-python@v6`
- `actions/setup-node@v6`
- `actions/upload-artifact@v4`
- `github/codeql-action/upload-sarif@v4`
- `codecov/codecov-action@v5`

Tag movement is a supply-chain risk. Pin each `uses:` to the full commit SHA (with a trailing `# vX.Y.Z` comment for readability), across all workflows for consistency.

Surfaced on PR #270 (the buff fix added `upload-artifact@v4` matching the file's existing convention); pinning is a focused, workflow-wide hardening better done in its own PR than mixed into a behavior fix. Fits the same supply-chain intent as #269.

ScienceIsNeato:
## Summary

Cleanup sweep covering three queued items:

### CI hardening (#271)
Every `uses:` across all three workflows is now pinned to a full commit SHA with a `# vX.Y.Z` comment: checkout v5.0.1, setup-python v6.2.0, setup-node v6.4.0, upload-artifact v4.6.2, codecov v5.5.5, codeql upload-sarif v4.36.2, and pypi-publish v1.14.0 (which replaces the moving `release/v1` branch ref). Tag movement can no longer silently change what runs in CI.

### Release provenance (#269)
The `publish-pypi` step now sets `attestations: true` explicitly. Trusted Publishing prerequisites (`environment: pypi`, `id-token: write`) were already in place, and the pinned pypi-publish v1.14.0 supports PEP 740 — so the next release's artifacts will carry verifiable provenance on PyPI.

### detect-secrets baseline fix (barnacle #253)
The gate's `--baseline` config-replay was structurally broken: the throwaway baseline carrying the real one's `plugins_used`/`filters_used` dropped detect-secrets' **default heuristic filters** (inflating findings ~4x in the reporting repo) and its re-scan results never matched the allowlist, so a valid committed `.secrets.baseline` suppressed nothing.

The fix exploits that `hashed_secret` is a sha1 of the secret value — **plugin-independent** — so:
- The scan always runs plain (default config, identical to what `detect-secrets scan > .secrets.baseline` uses), making `(path, hashed_secret)` pairs line up with the committed baseline
- The baseline's `results` are loaded read-only as the allowlist (existing mechanism, now actually effective)
- The baseline file itself is excluded from the walk and report — its recorded sha1 hashes trip the high-entropy detectors on every scan
- ~180 lines of replay machinery deleted (`_create_plugin_config_baseline`, `_load_tmp_baseline_report`, `_parse_detect_secrets_report`)

Also fixed the env-dependent `test_run_with_detect_secrets_failure` (failing locally whenever bandit/semgrep aren't importable: the mocked futures misaligned with the resolved scanner list).

## Test plan

- [x] 46 detect-secrets tests pass, including 3 new regression tests (never-replay, baseline-self-exclusion, walk pruning)
- [x] Full unit suite: 3177 passed; only the 5 pre-existing jinja2 env-dependent failures remain (down from 6)
- [x] Full swab clean

Closes #271, closes #269, closes #253.

This PR hardens CI, enables PyPI attestations, and fixes detect-secrets baseline handling.

- CI hardening: pins all GitHub Actions uses: entries to full commit SHAs (with version comments) across .github/workflows/release.yml, slopmop-sarif.yml, and slopmop.yml (checkout v5.0.1, setup-python v6.2.0, setup-node v6.4.0, upload-artifact v4.6.2, codecov v5.5.5, codeql upload-sarif v4.36.2, pypa/gh-action-pypi-publish v1.14.0). Added persist-credentials: false to checkout steps where relevant.
- Release provenance: publish-pypi step now sets attestations: true; trusted-publishing prerequisites (environment: pypi, id-token: write) were already present so future PyPI uploads will carry PEP 740-compatible provenance.
- detect-secrets baseline fix: removed ~180 lines of temp-baseline replay machinery and three helper methods; scans now run with default detect-secrets config, committed baseline results are used read-only as an allowlist, the baseline file is excluded from scanning/reporting, parsing uses result.stdout only and treats unparseable stdout as fail-closed, and an env-dependent scanner-import test was fixed.
- Tests: 46 detect-secrets tests pass (including 3 new regression tests); full unit suite: 3,177 passed (5 pre-existing jinja2 environment-dependent failures remain); swab clean.
- Closes: `#271`, `#269`, `#253`


ScienceIsNeato:
[fixed_in_code] Fixed in commit 6217ac3: persist-credentials: false added to all 9 checkout steps across release.yml, slopmop.yml, and slopmop-sarif.yml. Verified safe before applying: the only git push in any workflow (release-context job) authenticates via an explicit RELEASE_PR_TOKEN remote-URL rewrite, never via persisted checkout credentials, and the github-release job uses gh with GH_TOKEN env.

ScienceIsNeato:
[fixed_in_code] Fixed in commit 6217ac3: report is now parsed from result.stdout only, so stderr noise can't corrupt parsing. Also went one step further on your root concern: a successful scan whose stdout doesn't parse now FAILS CLOSED ('scan succeeded but stdout was not a parseable JSON report') instead of passing open — a corrupt report can no longer hide real secrets. Added regression test test_detect_secrets_parses_stdout_not_combined_output (stderr-polluted combined stream, clean stdout → passes) and updated the json-error test to assert fail-closed.

ScienceIsNeato:
[invalid_with_explanation] The doc is correct as written: tests/unit/test_version_consistency.py exists in this repo and is precisely the enforcement point — it imports scripts/sync_version.py and calls sync.check() against the live repo tree (line 60), failing CI on any drift. tests/unit/test_sync_version.py also exists but unit-tests the script's functions against tmp fixtures; it doesn't enforce repo state. The docstring in slopmop/_version.py names test_version_consistency.py as the enforcement test for the same reason.

PR #285, opened 2026-06-13
ci: dogfood the published slop-mop-action @v1

ScienceIsNeato:
## Summary

Adds a permanent advisory CI job that runs [`ScienceIsNeato/slop-mop-action@v1`](https://github.com/ScienceIsNeato/slop-mop-action/releases/tag/v1.0.0) against slop-mop itself — the same `@v1` ref consumers use. If a future slopmop release breaks the action's grade / SARIF / output contract, this job catches it here instead of in users' repos.

**Design choices**
- **Unpinned `@v1`** — intentional. The point is to validate what `@v1` resolves to today, i.e. the real consumer experience. (Our own action, internal dogfood; the supply-chain SHA-pinning from #277 was about third-party actions.)
- **Advisory** — `fail-on-failure: false`, so it doesn't double-gate the repo; `slopmop-sarif.yml` remains the enforcing gate. The job still goes **red if the action itself breaks**, which is the dogfood signal.
- **`upload-sarif: false`** — the primary gate already uploads the authoritative SARIF; this avoids duplicate Code Scanning alerts.
- Surfaces the hull grade in the job summary.

## Background

The action was validated end-to-end on a throwaway branch before tagging v1.0.0 (grade `B — serviceable`, all steps green, SARIF upload + outputs + minimum-grade enforcement confirmed against published slopmop 2.5.0). This makes that validation continuous.

## Summary

Adds a new GitHub Actions workflow (Action Dogfood) that runs the published ScienceIsNeato/slop-mop-action@v1 against this repository as an advisory dogfood check. Runs on pull_request, push to main, and workflow_dispatch. The job is advisory (fail-on-failure: false), disables SARIF upload, records the action outputs (grade, hull, provisional) via environment variables, appends the hull grade to GITHUB_STEP_SUMMARY, and exposes a grade/hull/provisional key-value line. The workflow uses contents: read permission, a full checkout (fetch-depth: 0) so git-aware gates resolve origin/main correctly, and a concurrency block (grouped by workflow+ref, cancel-in-progress: true) to cancel superseded runs.

## Changes

- .github/workflows/action-dogfood.yml: New workflow file (+55/-0)

### DIFF
--- a/.github/workflows/action-dogfood.yml
+++ b/.github/workflows/action-dogfood.yml
@@ -20,6 +20,10 @@ on:
 permissions:
   contents: read
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   dogfood:
     name: slop-mop-action @v1
@@ -28,6 +32,11 @@ jobs:
       - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd  # v5.0.1
         with:
           persist-credentials: false
+          # Full history: git-aware gates (diff coverage, gate-dodging)
+          # resolve origin/main, matching the primary scour workflow. A
+          # shallow clone makes them skip, which both diverges this grade
+          # from the real gate and inflates "provisional".
+          fetch-depth: 0
 
       - id: sm
         uses: ScienceIsNeato/slop-mop-action@v1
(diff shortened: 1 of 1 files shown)


ScienceIsNeato:
[fixed_in_code] Fixed in 6edee74: added fetch-depth: 0 to the checkout so git-aware gates (diff coverage's git diff origin/main...HEAD, gate-dodging) resolve origin/main, matching the primary scour workflow. This was also the cause of the provisional=true we observed — the shallow clone made those gates skip. Now the dogfood grade tracks the real gate.

ScienceIsNeato:
[fixed_in_code] Fixed in 6edee74: the report step now passes the action outputs (grade/hull/provisional) via env: and references $GRADE/$HULL/$PROVISIONAL in the run block, so values are treated as literal strings rather than expanded into the shell. Closes the zizmor template-injection finding.

ScienceIsNeato:
[fixed_in_code] Fixed in 6edee74: added a concurrency block (group by workflow+ref, cancel-in-progress: true) to cancel superseded runs. Closes the zizmor concurrency-limits finding.

PR #287, opened 2026-06-16
feat(hooks): guard commits against already-merged/deleted branches

ScienceIsNeato:
## What & why

An agent in [welcome-to-willville#47](https://github.com/ScienceIsNeato/welcome-to-willville/pull/47) added a pre-commit guard for a problem it framed as repo-local — but it's **universal to every slopmop user**: after a PR merges, it's easy to keep committing on the now-dead branch, and that work has to be redone on a fresh one.

slopmop already had a **pre-push** guard for the merged-PR case. This steals the thinking and ports the business logic into the **pre-commit** hook, so the dead branch is caught at *commit* time — before work piles up — and it handles more states than the push guard.

## The guard (ported from `check-branch-not-merged`, POSIX-sh)

Runs at the top of the generated pre-commit hook, before `sm swab`. Detection, strongest first — each **fail-open** so offline commits are never blocked:

1. A **merged PR** for the branch via `gh` (catches squash/rebase) — authoritative
2. The **remote head is gone** (`git ls-remote`) — the usual post-merge state
3. **HEAD already contained** in a moved-on default branch (ref math) — non-squash merge

**Allowed** without question: integration branches (`main`/`master`/`develop`), never-pushed branches, base-tracking branches (`checkout -b X origin/main`), and mid rebase/merge/cherry-pick. Override one commit with `git commit --no-verify`.

Key adaptation: since the guard runs *before* swab, allow-cases **fall through** (`return 0`) instead of `exit`, so they never skip the gates — only a real block exits non-zero.

## Tests

- Structural assertions (all three strategies + allow-lists + escape hatch present, guard ordered before swab)
- **POSIX-sh validation** of the generated hook across `sh`/`dash`/`bash -n`
- **Functional** tests against real git states using a local bare remote: blocks a branch merged into `origin/main`; allows fresh/open and integration branches

Verified manually across all 6 branch states (merged, deleted-remote-head, fresh, never-pushed, base-tracking, integration).

## Summary

Adds a commit-time (“dead-branch”) pre-commit hook guard that blocks commits on branches that are already merged/closed on GitHub or whose remote head has been deleted. Ported from prior logic and designed to be fail-open (offline-safe), acting as an earlier backstop than the existing push-time guard.

## Detection (strongest first, fail-open)

1. Merged PR detection for the current branch head via `gh pr list --state merged` using `--jq '.[0].number // empty'` (avoids `null` false-blocks).
2. Deleted remote head detection via `git ls-remote`.
3. “Already in default branch” detection via ref/merge-base containment.

## Allowed commits / bypasses

Explicitly allows: integration branches (`main`/`master`/`develop`), detached HEAD / never-pushed (no upstream), mid rebase/merge/cherry-pick, and base-tracking branches that track the default branch (e.g., created via `checkout -b X origin/<default>`) even if merged-PR data is stale. Provides a single-commit override via `git commit --no-verify`. Also fixes edge cases for renamed tracking branches so they aren’t incorrectly skipped.

## Changes

- `.willville.json`: Updated agent metadata to reflect the merged/dead-branch commit guard work.
- `slopmop/cli/hooks.py`: Added `_generate_merged_branch_guard()` and embedded it into the generated pre-commit hook; updated install/header messaging to document merged/deleted-branch commit refusal while keeping it fail-open.
- `tests/unit/test_git_hooks.py`: New suite validating guard generation/embedding, POSIX `sh -n`, strategy ordering/allow-lists, merged/deleted-remote-head behaviors (including via a local bare remote), jq empty-result handling, base-tracking ordering, renamed-branch behavior, and managed-hook parsing.
- `tests/unit/test_sm_cli.py`: Removed outdated unit coverage for internal hook helper behavior.
- `README.md`: Updated “Use with pre-commit” docs to describe the new pre-commit behavior and `git commit --no-verify` override.


ScienceIsNeato:
[fixed_in_code] Fixed in 6e7da64: changed the gh --jq filter to '.[0].number // empty'. On an empty merged-PR list jq's '.[0].number' emits the literal 'null' (non-empty → false-block as 'PR #null'); '// empty' yields an empty string so open branches pass. Verified with a fake gh that applies the --jq filter: empty result → allow, real merged PR → block. (Heads-up: the welcome-to-willville check-branch-not-merged original has this same bug.)

ScienceIsNeato:
[fixed_in_code] Fixed in 6e7da64: moved the base-tracking early-return ([ "$_sm_remote_branch" != "$_sm_branch" ] && return 0) ABOVE the gh check. A base-tracking branch (checkout -b X origin/main) tracks the default, not its own remote branch, so it can't be closed out — and running gh on it false-blocked when an older merged PR reused the local branch name. New test confirms a base-tracking branch with a stale merged PR (#99) is allowed.

ScienceIsNeato:
[fixed_in_code] Fixed in 3507f07: the discriminator now checks 'tracks the DEFAULT branch' ([ "$_sm_remote_branch" = "$_sm_default_branch" ]) instead of 'remote name != local name'. A renamed feature branch (git push -u origin local:remote) tracks its own non-default remote ref, so it's no longer skipped — and the gh/ls-remote checks now query $_sm_remote_branch (the real remote name). New test test_guard_checks_renamed_tracking_branch confirms a merged renamed branch is blocked, not skipped.

ScienceIsNeato:
[fixed_in_code] Fixed in 3507f07 (same change as the paired thread): narrowed the bypass to default-branch tracking only. A local alias tracking origin/feat/old is no longer treated as base-tracking — it tracks a non-default ref, so all three checks run, using the remote branch name. Verified by the new renamed-tracking-branch test (blocks on a stale merged PR rather than bypassing).

ScienceIsNeato:
[fixed_in_code] Fixed in 3507f07: added test_guard_blocks_deleted_remote_head — it pushes a branch, deletes the head directly in the bare remote (so the local tracking ref persists, the realistic post-merge state), and asserts the guard blocks via strategy #2 (git ls-remote sees the head gone). All three detectors now have direct coverage.

ScienceIsNeato:
[fixed_in_code] Fixed in 8840ed3: removed the redundant local 'import subprocess' from test_guard_checks_renamed_tracking_branch — subprocess is already imported at module level (test_git_hooks.py line 7).

PR #293, opened 2026-06-17
feat(gate): overconfidence:dangling-references — broken Markdown links (#289)

### LINKED ISSUE #289: feat(gate): overconfidence:dangling-references — links and assets that resolve in the agent's head, not on disk
### Problem

An agent writes a link, an ``, a `<script src>`, a `<link href>`, or an in-page `#anchor` pointing at a path that doesn't exist. The page renders fine — the broken reference only surfaces when someone clicks it (404) or the asset fails to load. The agent assumed the path resolved without checking.

This is **overconfidence**: *"it looks linked, therefore it works"* — silently breaks at runtime, the same failure shape as code that compiles but throws in production. The agent's mental model of the file tree disagreed with the actual file tree, and nothing in the loop made it look.

### Proposed solution

A deterministic, network-free gate that resolves every **internal** reference against the repo:

- relative/absolute internal `href`, `src`, `srcset`, `poster`, `<link href>` → must exist on disk (respecting the site root)
- in-page `#fragment` links → must match an element `id`/`name` on that page
- flag references to files outside the published tree

Internal-only by default; external/networked link-checking is flaky and belongs behind an opt-in. Prototype as a custom gate first per [New Gate Protocol](../DOCS/NEW_GATE_PROTOCOL.md).

### Alternatives considered

- **External link checkers** — rejected for the default path: networked, flaky, slow; not deterministic.
- **Runtime 404 monitoring / crawl after deploy** — rejected: that's after the slop has already shipped.

---
Part of a web-layer slop set: `overconfidence:access-blindness`, `overconfidence:dangling-references`, `deceptiveness:unfounded-metadata`, `myopia:conflicting-metadata`.

ScienceIsNeato:
Closes #289. A PURE, stack-agnostic gate that catches the "it looks linked, therefore it works" overconfidence — an agent writes a relative Markdown link or image, it renders fine, and moves on, but the target was renamed/moved and the reference is now dead.

## Scoped to be false-positive-free *by construction*, not by filtering

The minefield I flagged on the issue (framework routes, base-path rewrites, hashed build assets) is an **HTML-serving** phenomenon. Scoping to **Markdown only** removes the serving layer entirely — Markdown relative links have exactly one universal resolution semantics ("relative to the file on disk"), so there's nothing left to infer.

Every decision is a literal string test or a `path.exists()` — never inference. A target is **skipped, never flagged**, when it:
- has a URL scheme (`http:`, `mailto:`, `tel:`) or is protocol-relative (`//…`)
- contains template markers (`{{ }}`, `{% %}`, `${`) → SSG/templated
- is a pure `#fragment` (anchor checking needs slug inference — out of scope)
- is site-absolute (`/…`) → ambiguous repo-root vs site-root
- has a non-source extension (`.html`, `.js`) → could be a built artifact

What's left is only a literal relative path to a repo file. If it (or the directory it names) isn't on disk, it's unambiguously broken. A `../` target that escapes the repo is skipped via a containment guard.

The cost is deliberate **false-negatives** (broken HTML `src`, anchors, external 404s) — all need inference or the network, so they're correctly excluded.

## Why it clears the breadth bar
`is_applicable` = repo has ≥1 Markdown file → **stack-agnostic, not web-only.** Every repo with a README/docs is in scope, and broken relative doc links are one of the most common agent failures on a rename/move.

## Verification
- **Zero false positives on slop-mop's own 32 Markdown files** (9 real references, all resolved — non-vacuous).
- 22 unit tests including one per skip class (external URL, mailto, fragment, site-absolute, templated, non-source ext, protocol-relative, repo-escape).
- Full unit suite (3282) green; pyright/ruff/bandit clean; local grade A.
- PURE, DIAGNOSTIC, SCOUR; registered in all three places (registry, reasoning catalog, regenerated `DOCS/GATE_REASONING.md`).

## overconfidence:dangling-references Gate

This PR adds an `overconfidence:dangling-references` gate that deterministically (network-free) flags broken internal Markdown link/image references where the destination file doesn’t exist in the repo.

### Scope & behavior
- **Checks Markdown only** (`.md`, `.markdown`)
- **Detects** literal relative link/image targets and reference definitions like `[id]: target`
- **Resolves on disk** relative to the referencing Markdown file and repo root
- **Flags** missing targets and **repo-escaping paths** (e.g., `../../outside/...`)
- **Skips** non-checkable targets: URL schemes (`http:`, `mailto:`, `tel:`), `//` protocol-relative, template markers (`{{ }}`, `{% %}`, `${`), pure fragments (`#anchor`), site-absolute paths (`/...`), non-source extensions (e.g., `.html`, `.js`), and paths escaping the published tree

### Implementation & registration
- Implemented as `DanglingReferencesCheck` in `slopmop/checks/general/dangling_references.py`
- Registered in `slopmop/checks/__init__.py`
- Added reasoning entry in `DOCS/GATE_REASONING.md`
- Added built-in reasoning composition in `slopmop/checks/metadata.py`
- Updated `.willville.json`

### Key round-1 fixes
- **Balanced parentheses** in link destinations (prevents truncation false positives)
- **Top-down directory pruning** to avoid traversing excluded dirs (`node_modules`, `.git`, `vendor`)
- **Treat repo-escaping targets as broken** (instead of passing)

### Tests
- Added `tests/unit/test_dangling_references_check.py` with coverage for metadata/applicability, failure modes, correct resolutions (including parentheses), and extensive “no false positives” cases.

### DIFF
--- a/.willville.json
+++ b/.willville.json
@@ -1,7 +1,7 @@
 {
   "agent": {
-    "status": "Made the URL-consistency gate a ship-time check instead of every-commit, and locked in that onboarding turns it on only for sites with real pages",
-    "direction": "Push and let CI confirm before the merge call",
-    "last_update": "2026-06-16T00:00:00Z"
+    "status": "Built a gate that catches broken doc links — markdown links pointing at files that aren't there after a rename or move",
+    "direction": "Kept it dead simple so it never cries wolf; opening the PR next",
+    "last_update": "2026-06-17T00:00:00Z"
   }
 }
--- a/DOCS/GATE_REASONING.md
+++ b/DOCS/GATE_REASONING.md
@@ -22,6 +22,12 @@ This file is generated from built-in gate metadata. Edit the gate reasoning sour
 - Tradeoffs: Coverage work slows down spikes and sometimes forces harness cleanup before the feature feels done.
 - Override When: Bend this for short-lived spikes, active incidents, or other explicitly time-critical work with agreement that the coverage debt gets paid back.
 
+### `overconfidence:dangling-references`
+
+- Rationale: An agent writes a relative Markdown link or image and moves on; the file renders, so it looks done.  The broken reference only surfaces when a reader clicks it — the agent's mental model of the file tree disagreed with the real one and nothing made it look.
+- Tradeoffs: Deliberately narrow to stay false-positive-free: only literal relative paths in Markdown are checked, so broken HTML src, anchors, and external links are missed by design.
+- Override When: Suppress for generated or vendored docs you do not author.  Add the directory to exclude_dirs in .sb_config.json.
+
 ### `overconfidence:missing-annotations.dart`
 
 - Rationale: Missing Dart annotations turn interfaces into vibes and push type noise downstream for somebody else to untangle.
--- a/slopmop/checks/__init__.py
+++ b/slopmop/checks/__init__.py
@@ -83,6 +83,7 @@ def _register_dart_checks(registry: CheckRegistry) -> None:
 def _register_crosscutting_checks(registry: CheckRegistry) -> None:
     """Register security, quality, and general checks."""
     from slopmop.checks.general.conflicting_metadata import ConflictingMetadataCheck
+    from slopmop.checks.general.dangling_references import DanglingReferencesCheck
     from slopmop.checks.general.interactive_assumptions import (
         InteractiveAssumptionsCheck,
     )
@@ -113,6 +114,7 @@ def _register_crosscutting_checks(registry: CheckRegistry) -> None:
     registry.register(DebuggerArtifactsCheck)
     registry.register(GateDodgingCheck)
     registry.register(ConflictingMetadataCheck)
+    registry.register(DanglingReferencesCheck)
     registry.register(InteractiveAssumptionsCheck)
     registry.register(RepeatedCodeCheck)
     registry.register(StringDuplicationCheck)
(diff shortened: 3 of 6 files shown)


ScienceIsNeato:
[fixed_in_code] Fixed in 5390818: _INLINE_RE now matches balanced parentheses ((?:[^()\s]+|\([^)]*\))+), so ./docs/api(v2).md is captured whole instead of truncating at the first ). Added test_existing_link_with_parentheses_in_path. Real FP — thanks for catching it.

ScienceIsNeato:
[fixed_in_code] Fixed in 5390818: added test_existing_link_with_parentheses_in_path under TestResolves, exactly as suggested.

ScienceIsNeato:
[fixed_in_code] Fixed in 5390818: _iter_markdown_files now uses os.walk(topdown=True) with in-place dirnames[:] pruning via should_prune_dir + is_path_excluded, so node_modules/.git/vendor are never descended into. Added test_noise_and_excluded_dirs_are_pruned.

ScienceIsNeato:
[fixed_in_code] Fixed in 5390818: _resolves now returns False (flag) when a target escapes the repo root — it can't resolve in any published/CI context, matching the issue's 'flag references outside the published tree' goal, and we no longer stat outside the repo. Internal ../ links that stay in-tree still resolve. Added test_repo_escaping_target_is_flagged + test_repo_internal_parent_link_resolves.

PR #297, opened 2026-06-19
ci(release): auto-bump slop-mop-action's slopmop pin on release

ScienceIsNeato:
Completes the "tie the pin to the release flow" plan. The companion change in [slop-mop-action#2](https://github.com/ScienceIsNeato/slop-mop-action/pull/2) pins `slopmop-version` by default (so the install cache is deterministic); this keeps that pin current automatically.

## What it does
A `bump-action-pin` job runs after the GitHub release is published and opens a PR on `ScienceIsNeato/slop-mop-action` setting its `slopmop-version` default to the just-released version.

- **Precise edit:** a range-bounded `sed` updates only the `slopmop-version` input's default — verified locally that every other input's `default:` is untouched and the result is valid YAML.
- **Reviewable:** opens a PR, never pushes to the action's main.
- **Best-effort, never blocks a release:** `continue-on-error` plus guards. By the time this job runs the release is already on PyPI and GitHub, so a missing/again-scoped `RELEASE_PR_TOKEN` or a clone failure just logs a warning.

## Dependency to be aware of
The job needs `RELEASE_PR_TOKEN` to have **write/PR access to `slop-mop-action`** (it already creates the release PR in this repo; this adds a second repo). If the token is scoped to slop-mop only, the job will warn and no-op — you'd bump the action manually that one time, or widen the token scope.

## Sequencing
Merge this before cutting 2.6.0 so that release auto-bumps the action from `==2.5.0` to `==2.6.0`.

Workflow YAML validated; GitHub-Actions hygiene gate passes (the new job uses only `run:` steps, no actions to pin).

## Automatic Version Pinning for `slop-mop-action`

Adds a `bump-action-pin` job to the release workflow (runs after `github-release` when `release-context.outputs.should_release` is `true`) to open a reviewable PR on `ScienceIsNeato/slop-mop-action` updating the `action.yml` input default `slopmop-version` to the newly released `VERSION`.

The job uses a bounded `sed` edit to update only the `slopmop-version` default (preserving the rest of the YAML), and is best-effort: it no-ops with warnings if the token is missing/insufficient, cloning fails, the edit produces no diff, or pushing fails. It creates/updates `chore/bump-slopmop-${VERSION}`, pushes the branch, and then attempts `gh pr create` against `main` (PR-creation failures are tolerated). The job is marked `continue-on-error` so release failures aren’t blocked.

Error-handling improvements from review:
- Explicit `git push` failure detection with a targeted warning (e.g., likely token write access on the target repo) instead of masking the root cause behind `gh pr create` errors.
- Fix for reruns with shallow clones by fetching the remote tracking ref before pushing so `--force-with-lease` behaves correctly and diagnostics remain accurate.

**Files changed:**
- `.github/workflows/release.yml`: Added `bump-action-pin` job (+66 lines)
- `.willville.json`: Updated agent release metadata (+2/-2 lines)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -468,3 +468,60 @@ jobs:
             --target "$TARGET" \
             --title "$TAG" \
             --notes-file release_body.md
+
+  bump-action-pin:
+    name: Bump slop-mop-action pin
+    needs:
+      - release-context
+      - github-release
+    if: needs.release-context.outputs.should_release == 'true'
+    runs-on: ubuntu-latest
+    steps:
+      # Keep the published slop-mop-action pinned to the freshly released
+      # version so its install cache stays deterministic and current. Opens a
+      # PR (reviewable) rather than pushing to main. Best-effort: the release
+      # is already published, so a failure here never blocks shipping.
+      - name: Open PR to bump the action's slopmop-version pin
+        continue-on-error: true
+        env:
+          GH_TOKEN: ${{ secrets.RELEASE_PR_TOKEN }}
+          VERSION: ${{ needs.release-context.outputs.version }}
+          ACTION_REPO: ScienceIsNeato/slop-mop-action
+        run: |
+          set -uo pipefail
+          if [ -z "${GH_TOKEN:-}" ]; then
+            echo "::warning::RELEASE_PR_TOKEN not set — bump slopmop-version to ==${VERSION} in ${ACTION_REPO} manually."
+            exit 0
+          fi
+
+          workdir="$(mktemp -d)"
+          if ! git clone --depth 1 \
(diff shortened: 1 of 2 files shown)


ScienceIsNeato:
[fixed_in_code] Fixed in ed55270: git push is now checked explicitly — on failure it emits a specific 'failed to push (token may lack write access)' warning and exits, so it no longer falls through to gh pr create. The pr-create fallback message now only covers the genuine already-exists / missing-PR-scope case.

ScienceIsNeato:
[fixed_in_code] Fixed in f52b708: added 'git fetch --depth 1 origin refs/heads/$branch:refs/remotes/origin/$branch || true' before the push so --force-with-lease has a tracking ref on reruns (no-op on first run). Reruns now push cleanly instead of failing and misreporting a token issue.

PR #298, opened 2026-06-19
feat(hooks): run sm scour at pre-push to catch scour-only failures locally

ScienceIsNeato:
\"I keep getting to CI only to find the scour failed.\" This PR has two parts:

## Part 1: scour on push

The native pre-push hook (`sm commit-hooks install`) now runs the full `sm scour` after the merged-branch guard — the same validation CI runs — so scour-only failures surface before the push, not on a red build.

### Why it's fast (the key concern)
The pre-commit swab already writes per-gate results to the cache, keyed by a per-gate fingerprint. The push scour **reuses** those, so only scour-only and changed gates run fresh. Measured on a sample repo:

| | wall-clock |
|---|---|
| cached scour (after swab) | **135 ms** |
| `--no-cache` scour | 3,647 ms |

≈ **27× faster**. Verified the full cycle: first scour after swab showed `6/38 cached` (swab gates reused), a second scour caches everything, and editing a file invalidates only the gates that touch it.

## Part 2: machine-wide install (`--global`)

Native git hooks are per-clone, not version-controlled — you'd have to re-run `sm commit-hooks install` in every repo. The `--global` flag fixes that:

```
sm commit-hooks install --global   # wires every repo on this machine
sm commit-hooks uninstall --global # removes it
```

Under the hood this writes hooks to `~/.slopmop/git-hooks/` and sets `git config --global core.hooksPath` to point there. The global hooks:
- **Delegate first** — if the repo already has a `.git/hooks/<name>` that isn't slopmop-managed, it runs that first so other tools' hooks are preserved.
- **Skip non-onboarded repos** — repos without `.sb_config.json` or `[tool.slopmop]` in `pyproject.toml` get `exit 0`; slopmop never touches them.
- **Prevent double-runs** — the delegation check skips slopmop-managed local hooks, so if you also have a per-repo install the hook doesn't fire twice.
- **Capture stdin correctly** — the pre-push variant slurps stdin once and feeds it to both the delegated hook and the merged-branch guard.

Both actions work outside a git repo (they target `~/.slopmop/git-hooks/`, not `.git/hooks/`).

## Behavior (pre-push hook)
- Merged-branch guard runs first (its stdin loop finishes before scour starts, so scour never consumes the pushed refs).
- A failing scour blocks the push with the structured results path and `git push --no-verify` to bypass once.
- Brings the **native git hooks to parity** with the pre-commit-framework path, which already ran `slopmop-scour` at the `pre-push` stage.

Generated POSIX-sh validated (`sh -n`); full suite (3300) green.

## Summary

Native `pre-push` now runs the full `sm scour` validation after the merged/deleted-branch guard, reusing pre-commit swab cache so scour-only/changed gates are fast. Cached scour is ~135ms vs ~3,647ms uncached (~27×). A failing scour blocks pushes (one-time bypass: `git push --no-verify`). Behavior is aligned with the pre-commit framework.

Adds machine-wide installation via `sm commit-hooks install --global` / `uninstall --global`, using git `core.hooksPath` with global hook delegation to repo-local hooks for onboarded repos (skipping slopmop for non-onboarded repos). Includes safety/fixups: passthrough delegation for all hook types, worktree-safe hook paths, warning before overwriting non-slopmop `core.hooksPath`, friendlier global `git config` errors, and POSIX `sh -n` validation using race-free temp files.

## Changes

- `slopmop/cli/hooks.py`: refactors `pre-push` (guard → cached `sm scour`, writes `.slopmop/last_scour.json`) and implements `--global` hook install/uninstall with delegation/preamble.
- `slopmop/sm.py`: adds `commit-hooks install --global` and `commit-hooks uninstall --global`.
- `README.md`: updates documented commit/push flow, cache behavior, guard behavior, and `--no-verify`.
- `tests/unit/test_git_hooks.py`: adds `sh -n` validation, asserts guard→scour ordering and push-block messaging, and adds global hook generation/delegation/install-uninstall coverage.
- `tests/unit/test_sm_cli.py`: updates expected generated `pre-push` script content.
- `.willville.json`: updates agent metadata/timestamps.

All **3,300** tests pass; generated POSIX shell scripts are validated with `sh -n`.

### DIFF
--- a/.willville.json
+++ b/.willville.json
@@ -1,7 +1,7 @@
 {
   "agent": {
-    "status": "Added --global flag to commit-hooks install so it wires up machine-wide git hooks via core.hooksPath — PR #298",
-    "direction": "Get PR #298 reviewed and merge, then dogfood with sm commit-hooks install --global",
-    "last_update": "2026-06-19T12:00:00Z"
+    "status": "PR #298 clean — scour on push + --global install, all CI green and review resolved",
+    "direction": "Ready to merge; after merge, dogfood with sm commit-hooks install --global",
+    "last_update": "2026-06-19T14:08:00Z"
   }
 }
--- a/.willville.json
+++ b/.willville.json
@@ -1,7 +1,7 @@
 {
   "agent": {
-    "status": "Wiring releases so they automatically keep the GitHub Action pointed at the just-shipped version — no more manual bumps, no more stale install caches",
-    "direction": "Land this so the next release bumps the action on its own",
-    "last_update": "2026-06-17T00:00:00Z"
+    "status": "Making the push hook run the full check suite locally, so people stop discovering in CI that something failed — and it reuses the quick-check results so it stays fast",
+    "direction": "Land it, keep an eye on whether the caching really saves the repeat work",
+    "last_update": "2026-06-19T00:00:00Z"
   }
 }
--- a/.willville.json
+++ b/.willville.json
@@ -1,7 +1,7 @@
 {
   "agent": {
-    "status": "Making the push hook run the full check suite locally, so people stop discovering in CI that something failed — and it reuses the quick-check results so it stays fast",
-    "direction": "Land it, keep an eye on whether the caching really saves the repeat work",
-    "last_update": "2026-06-19T00:00:00Z"
+    "status": "Added --global flag to commit-hooks install so it wires up machine-wide git hooks via core.hooksPath — PR #298",
+    "direction": "Get PR #298 reviewed and merge, then dogfood with sm commit-hooks install --global",
+    "last_update": "2026-06-19T12:00:00Z"
   }
 }
--- a/README.md
+++ b/README.md
@@ -130,8 +130,14 @@ adopted slop-mop: in a repo that hasn't been onboarded (`sm init` +
 blocking the commit.
 
 If you prefer raw git hooks without the pre-commit framework,
-`sm commit-hooks install` does the same job with sm-managed scripts. Those
-hooks also carry a **merged/deleted-branch guard**: the pre-commit hook
+`sm commit-hooks install` does the same job with sm-managed scripts: swab on
+every commit, and the full **`sm scour`** on push — the same validation CI
+runs, so scour-only failures are caught before you push instead of on a red
+build. The push scour reuses the swab results from cache (keyed per gate), so
+only scour-only and changed gates run fresh; bypass once with
+`git push --no-verify`.
+
+Those hooks also carry a **merged/deleted-branch guard**: the pre-commit hook
 refuses a commit when the current branch has already been merged (a merged
 PR, the remote head deleted, or HEAD already contained in a moved-on default
 branch) so you don't pile work onto a branch that's closed out and would have
--- a/slopmop/cli/hooks.py
+++ b/slopmop/cli/hooks.py
@@ -434,6 +434,17 @@ def _hooks_status(project_root: Path, hooks_dir: Path) -> int:
 
     print_project_header(str(project_root))
     print(f"📁 Hooks dir: {hooks_dir}")
+
+    # Surface the global install state so the user knows which hooks are active.
+    global_dir = _global_hooks_dir()
+    current_hooks_path = subprocess.run(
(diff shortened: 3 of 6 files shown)


ScienceIsNeato:
[fixed_in_code] Fixed in cf86b92: before setting core.hooksPath we now read the existing value and print a warning if we're about to replace a non-slopmop path. The uninstall path already only unsets when the value matches our dir (unchanged).

ScienceIsNeato:
[fixed_in_code] Fixed in cf86b92: the global preamble now uses `$(git rev-parse --git-dir 2>/dev/null)/hooks/{hook_name}` instead of the hardcoded `$_sm_root/.git/hooks/` path, so linked worktrees resolve the correct hooks directory.

ScienceIsNeato:
[fixed_in_code] Fixed in cf86b92: wrapped the git config --global call in try/except subprocess.CalledProcessError with a user-friendly error message and return 1.

ScienceIsNeato:
[fixed_in_code] Fixed in cf86b92: _is_posix_sh now uses tempfile.NamedTemporaryFile(delete=False) with explicit p.unlink(missing_ok=True) in a finally block, eliminating the TOCTOU race and the temp file leak.

ScienceIsNeato:
[fixed_in_code] Fixed in 83afa1f: _hooks_install --global now writes passthrough delegation scripts for all client-side hook types (commit-msg, prepare-commit-msg, post-checkout, post-commit, etc.) via the new _GLOBAL_PASSTHROUGH_HOOKS list and _generate_passthrough_hook(). Each script forwards to $(git rev-parse --git-dir)/hooks/<name> if executable. _hooks_uninstall removes them on uninstall.

ScienceIsNeato:
[out_of_scope_ticketed] Valid finding. Tracked as #299. The fix requires park_slopmop_hook to detect whether core.hooksPath points at our global dir and park the global hook file instead. Non-trivial and orthogonal to this PR's main feature; filed to address separately.

ScienceIsNeato:
[invalid_with_explanation] The delegation check (! grep -q '# MANAGED BY SLOP-MOP' $_sm_local) already prevents double-running when the local hook is slopmop-managed. The 'duplicate' scenario requires both the global install AND the pre-commit framework configured to run slopmop tools, which is an unusual combination — most users use one or the other. The output is idempotent (same gate runs produce the same results), so duplicate runs don't cause incorrect behavior, just slightly slower commits for this niche setup.

ScienceIsNeato:
[invalid_with_explanation] The per-repo hooks (sm commit-hooks install without --global) also check only for .sb_config.json or [tool.slopmop] and don't do a further .slopmop/ existence check. sm swab and sm scour handle uninitialized repos gracefully — they surface an error rather than hanging. Making the global check stricter than the per-repo check would create a confusing inconsistency. The init flow naturally writes .sb_config.json as part of onboarding, so a repo in that state has chosen slopmop.

ScienceIsNeato:
[fixed_in_code] Fixed in cfccddf: the test now loops over _GLOBAL_PASSTHROUGH_HOOKS and asserts every hook file exists in the global dir.

ScienceIsNeato:
[fixed_in_code] Fixed in cfccddf: changed to ['pre-commit', 'pre-push', *_GLOBAL_PASSTHROUGH_HOOKS] per the Ruff RUF005 suggestion.

ScienceIsNeato:
[fixed_in_code] Fixed in cfccddf: assertion now checks the full '.git/hooks/{hook_name}' form, which would catch a regression to .git/hooks/...

ScienceIsNeato:
[fixed_in_code] Fixed in 8193e2a: _hooks_install (non-global path) now checks if core.hooksPath already points at our global dir and prints a clear warning that per-repo hooks written to .git/hooks won't run while global hooks shadow them, with a note on how to update or uninstall the global hooks.

ScienceIsNeato:
[fixed_in_code] Fixed in 8193e2a: _hooks_status now calls git config --global --get core.hooksPath and, if it points at our global dir, reports it clearly above the per-repo section so the user knows which hooks git will actually run.

PR #302, opened 2026-06-19
feat(sail): make sm sail the autonomous drive-to-green verb; collapse buff under it

ScienceIsNeato:
## Why

Agents kept ending a turn with CI mid-flight, a red check, or unanswered review comments — so the human had to repeatedly say "buff the pr." The fix is one verb. `sm sail` was *already* a state-machine dispatcher that wraps swab/scour/buff, but it ran **one step per call** and the docs steered agents at `buff` directly. This PR makes `sail` drive on its own and collapses the agent-facing surface onto it.

## What changed

**`sm sail` now drives autonomously (`slopmop/cli/sail.py`).** A single invocation keeps dispatching the next workflow step (swab → scour → `buff watch` → review triage) until the workflow **parks** on something only the agent/human can do — fix a failing gate, commit, push, open the PR, resolve threads — or reaches PR_READY. Parking is detected by the persisted state failing to advance (tool steps move it; informational handlers don't). A cycle guard + step cap prevent spinning.

- **Safety preserved:** commit, push, and PR creation stay agent steps (they need an authored message/body), so sail still never mutates git or publishes on its own. It just stops making the agent re-invoke it between every validation step.
- **Backward compatible:** the existing per-state dispatch tests pass unchanged (a pinned state parks after one step). Adds `TestSailDrive` covering chaining, the PR_READY terminal stop, immediate parking, and the cycle guard.

**Docs recentered on one verb:**
- `/sm-sail` is now the primary drive-to-green loop — run sail, do the one thing it parks on, repeat until "PR ready." Carries the honesty rule, the two sanctioned early-returns (decision fork / unclearable blocker), and convergence guidance.
- `/sm-buff` demoted to the under-the-hood triage reference (the verbs sail runs for you; surgical use only).
- `core.md`: "The loop" + "Definition of done" recenter on sail ("not done until sail says PR ready"); substitution table routes the PR-driving impulses to `sm sail`.

## Notes
- Supersedes the earlier framing of this PR (a standalone `/sm-buff` loop). buff is now strictly under sail's hood.
- Also folds a 3-line CHANGELOG wording fix ("every repo" → "every onboarded repo") that #300 merged before it landed.
- The new sail behavior ships here; it's unit-tested but not yet dogfoofed against a live PR (the locally-installed `sm` is the released 2.5.0).

### DIFF
--- a/.willville.json
+++ b/.willville.json
@@ -1,7 +1,7 @@
 {
   "agent": {
-    "status": "Made the post-push helper drive a PR all the way to green on its own, instead of stopping halfway — PR #302, clean and ready",
-    "direction": "So agents stop handing back half-finished PRs; folded the same 'not done until it's green' rule into the shared instructions",
-    "last_update": "2026-06-19T16:16:00Z"
+    "status": "Made `sm sail` drive a change all the way to a green PR on its own, so there's one command to run instead of a handful — PR #302, green",
+    "direction": "Point everyone at that single command and tuck the rest of the tooling behind it",
+    "last_update": "2026-06-19T22:01:00Z"
   }
 }
--- a/.willville.json
+++ b/.willville.json
@@ -1,7 +1,7 @@
 {
   "agent": {
-    "status": "PR #298 clean — scour on push + --global install, all CI green and review resolved",
-    "direction": "Ready to merge; after merge, dogfood with sm commit-hooks install --global",
-    "last_update": "2026-06-19T14:08:00Z"
+    "status": "Made the post-push helper drive a PR all the way to green on its own, instead of stopping halfway — PR #302, clean and ready",
+    "direction": "So agents stop handing back half-finished PRs; folded the same 'not done until it's green' rule into the shared instructions",
+    "last_update": "2026-06-19T16:16:00Z"
   }
 }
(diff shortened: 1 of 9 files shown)


ScienceIsNeato:
[no_longer_applicable] The 2.7.0 CHANGELOG entry was rebased out of this PR (#302 is now docs-only); it lives in #300. The 'every repo' wording is a valid catch and is being fixed there instead.

ScienceIsNeato:
[fixed_in_code] Fixed in 670e6fc: workflow fence labeled text, command-example fence labeled bash in commands/sm-buff.md.

ScienceIsNeato:
[fixed_in_code] Fixed in 670e6fc: the sm-buff-watch example fence in core.md is now labeled text.

ScienceIsNeato:
[fixed_in_code] Fixed in 670e6fc: both fences in the claude template labeled (text workflow, bash command).

ScienceIsNeato:
[fixed_in_code] Fixed in 670e6fc: single-quoted slopmop[all] in commands/sm-buff.md so the shell does not glob-expand the brackets.

ScienceIsNeato:
[fixed_in_code] Fixed in 670e6fc: single-quoted slopmop[all] in the claude template too.

ScienceIsNeato:
Rebased onto current main (post-#300 merge) and folded in a 3-line CHANGELOG wording fix ("every repo" → "every onboarded repo") — the catch from review thread 1, which #300 merged before it landed. Bundled here rather than spawning a separate PR to avoid an extra CI cycle.

ScienceIsNeato:
[fixed_in_code] Fixed in bdf6561: reworked the section to present two distinct sanctioned early-returns — a decision you can't make (needs_human_feedback) and a blocker you can't clear (circuit breaker) — instead of calling needs_human_feedback the sole exit. They no longer contradict.

PR #305, opened 2026-06-20
feat(gates): external-dependency contract + actionlint gate (doctor manifest groundwork)

ScienceIsNeato:
First step of the "doctor owns what scour needs" plan. Gates declare their external dependencies in a structured, versioned contract so a single authority (`sm doctor`, then the GitHub Action) can enumerate and install them — replacing the scattered `shutil.which` detection that makes "what a gate needs" implicit knowledge.

**Gates-first, by design.** Doctor changes are mechanical *if* gates can speak for themselves; this lands the contract + converts one gate to prove the shape before touching the other 40.

## The contract (`slopmop/checks/base.py`)

- **`Requirement` / `Requirements`** dataclasses. `version` is an **exact pin**, never a floor — a scanner's version is part of the gate definition (bumping bandit changes findings), so it's bumped via a reviewed change, not floated at install time. We explicitly walked back auto-upgrade for this reason.
- **`REQUIREMENTS_MANIFEST_SCHEMA_VERSION` + `build_requirements_document()`** — the manifest is a *consumed contract* (doctor emits it, the Action installs from it), so it's schema-versioned from day one. `to_manifest()` sorts by `(kind, name)` for **byte-stable, cacheable** output (the constraint that keeps Action spend bounded — the Action can cache-key on the manifest).
- **`BaseCheck.requirements()`** (default empty), **`missing_requirements()`** (venv-aware `find_tool`, honours `alternatives`), and **`requirement_block_result()`** — the **three-state model**:
  - required tool missing → **`ERROR`** (could-not-run; fails the verdict — `all_passed` counts errors — so a broken environment can't silently pass a merge gate)
  - optional tool missing → does **not** block
  - not-applicable → unchanged

  This directly fixes the silent-decay risk: today the actionlint path returns `PASSED` when actionlint is absent. The mechanism to make a *required* missing tool non-green now exists.

## First consumer (`github_actions.py`)

The actionlint gate declares actionlint as an **optional, config-gated** requirement and resolves it through **one** `find_tool` path instead of two inline `shutil.which` calls — so "what the gate looks for" and "what it declares" can't drift. Behavior-preserving (actionlint stays optional), but now visible to doctor.

## Tests
`tests/unit/test_gate_requirements.py` — default-empty, config-gated declaration, missing/present/alternatives detection, **required→ERROR vs optional→no-block**, and manifest **determinism + schema version**. Full suite green (3320).

## Follow-ups (not here)
1. Convert the hard cases (`security` multi-scanner, `duplicate_strings` fallback chain) — they shape the API.
2. Convert the rest by category; wire `requirement_block_result` into required-tool gates' `run()`.
3. Registry enforcement test (every gate overrides `requirements()`) — **must land last**, after all 41 are converted.
4. `sm doctor --required-deps` manifest emitter → then the **v2** Action consumer.

## Summary

Introduces a versioned external-dependency contract for gates (for future `sm doctor` / Action integration) and updates the first consumer: the GitHub Actions hygiene gate’s `actionlint` handling.

### Key changes
- **Dependency contract (`slopmop/checks/base.py`)**: Adds `Requirement`/`Requirements` dataclasses with **exact pinned versions** and a **schema-versioned, byte-stable** `build_requirements_document()` manifest (stable ordering by `(kind, name)` and **sorted `alternatives`** for determinism/caching).
- **Three-state dependency gating**: `BaseCheck` adds `requirements()` (default empty), `missing_requirements(project_root)` (system tools resolved locally with alternatives), and `requirement_block_result()` implementing **missing required → `ERROR` (fail)** vs **missing optional → no-block**.
- **First consumer (`github_actions.py`)**: Declares **`actionlint` as optional and config-gated**; resolves the executable **once per run** via `find_tool()` and threads it into `_actionlint_findings()`. The pass note distinguishes:
  - “actionlint disabled” (config off)
  - “actionlint checked” (resolved)
  - “actionlint not installed” (optional missing)

### Tests
- Adds `tests/unit/test_gate_requirements.py` for default/config-gating, missing/present/alternatives detection, required vs optional semantics, and manifest determinism/schema versioning.
- Updates `tests/unit/test_github_actions_hygiene.py` to validate resolved/disabled/None behavior and the three-state pass note wording (no `shutil.which` patching).
- Full suite passes (**3320 tests**).

### DIFF
--- a/slopmop/checks/base.py
+++ b/slopmop/checks/base.py
@@ -582,7 +582,7 @@ class Requirement:
             "name": self.name,
             "version": self.version,
             "optional": self.optional,
-            "alternatives": list(self.alternatives),
+            "alternatives": sorted(self.alternatives),
             "reason": self.reason,
         }
 
--- a/slopmop/checks/base.py
+++ b/slopmop/checks/base.py
@@ -531,6 +531,86 @@ STANDARD_CONFIG_FIELDS = [
 ]
 
 
+# ── Gate external-dependency contract ────────────────────────────────
+#
+# Gates that shell out to external tools declare what they need here so a
+# single authority — ``sm doctor`` and, downstream, the GitHub Action — can
+# enumerate and install dependencies from one source of truth instead of each
+# gate detecting tools ad hoc (the scattered ``shutil.which`` problem).
+#
+# The manifest these produce is a CONSUMED CONTRACT (doctor reads it; the
+# Action installs from it), so its shape is versioned. Bump the schema version
+# on any breaking change to the serialized field set.
+REQUIREMENTS_MANIFEST_SCHEMA_VERSION = 1
+
+
+@dataclass(frozen=True)
+class Requirement:
+    """A single external dependency an enabled gate needs to run.
+
+    ``optional`` encodes the distinction that matters for a CI merge gate:
+
+    - **required** (``optional=False``) — the gate *cannot run* without it.
+      A missing required tool must surface a visible, non-green
+      "could-not-run" state, never a silent pass. (See
+      :meth:`BaseCheck.requirement_block_result`.)
+    - **optional** (``optional=True``) — the gate degrades but still runs;
+      a missing optional tool is reported (so doctor can recommend it) but
+      does not block.
+
+    ``version`` is an EXACT pin (e.g. ``"1.7.5"``), never a floor. The pin is
+    part of the gate definition — bumping a scanner can change findings — so it
+    is bumped deliberately via a reviewed change, not floated at install time.
+    ``None`` means "any present version is acceptable" (e.g. system ``git``).
+
+    ``alternatives`` lists interchangeable names that also satisfy the
+    requirement (any one present is enough) — for tools resolvable under more
+    than one binary name / install path.
+    """
+
+    kind: str  # "system" | "python" | "npm" | "env"
+    name: str
+    version: Optional[str] = None  # exact pin; None = any present version
+    reason: str = ""
+    optional: bool = False
+    alternatives: tuple[str, ...] = ()
+
+    def to_manifest(self) -> Dict[str, Any]:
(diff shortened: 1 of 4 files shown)


ScienceIsNeato:
[fixed_in_code] Fixed in 4b2bff8: to_manifest() now returns sorted(self.alternatives) so manifest output is byte-stable regardless of declaration order — closing the determinism hole you flagged.

ScienceIsNeato:
[fixed_in_code] Fixed in 4b2bff8: to_manifest() now returns sorted(self.alternatives), so manifest output is byte-stable regardless of declaration order — closing the determinism hole.

ScienceIsNeato:
[fixed_in_code] Fixed in 4b2bff8: the actionlint gate resolves the executable once in run() and threads the path into _actionlint_findings(), eliminating the per-workflow-file re-discovery.

ScienceIsNeato:
[fixed_in_code] Fixed in this push: the pass note is now three-way — 'actionlint disabled' when run_actionlint is off, 'actionlint checked' when resolved, 'actionlint not installed' otherwise. Added test_passed_note_distinguishes_disabled_from_missing.

ScienceIsNeato:
[fixed_in_code] Fixed: test_passed_note_distinguishes_all_three_states now asserts all three branches — disabled, checked (find_tool mocked to a path + clean runner), and not installed (find_tool mocked to None).

ScienceIsNeato:
[fixed_in_code] Fixed: combined the two context managers into a single with-statement (SIM117). Ruff now clean on the file.

ScienceIsNeato:
[fixed_in_code] Fixed: extracted BaseCheck.resolve_requirement_path() (name + alternatives) as the single resolution path. Both missing_requirements and the gate's _actionlint_path now go through it, so a declared alternative can't make doctor report 'present' while the gate reports 'not installed'. Added a regression test asserting the two views agree.

PR #306, opened 2026-06-20
feat(gates): convert security + duplicate-strings (hard-case requirements)

ScienceIsNeato:
**Stacked on #305** (base = `feat/gate-requirements-contract`). Converts the two gates we flagged as the API-stressing cases — if the contract shape handles these, the remaining ~38 are mechanical.

## What they stressed (that actionlint didn't)

- **install channel ≠ detection method.** `semgrep` is pip-installed but invoked as a binary. Added `probe` (binary | import | env | none), orthogonal to `kind`. semgrep = `kind=python, probe=binary`.
- **install name ≠ import name.** `detect-secrets` imports as `detect_secrets`. Added `import_name`.
- **python-package probing.** `BaseCheck.is_requirement_satisfied()` + `_module_available()` (via `find_spec`, dodging stevedore's plugin-enumeration warnings — the same reason the security gate already used `find_spec`). `missing_requirements` now probes every kind correctly, not just system binaries.

Both new fields serialize into the manifest. Still **schema v1** — it's unreleased with no consumers, so evolving it pre-consumer is legitimate; the version exists to protect consumers, of which there are none until the doctor emitter lands.

## The conversions

- **`security:local`** declares its configured scanners (bandit/semgrep/detect-secrets) as optional requirements and replaces `_is_scanner_available`'s scattered `find_spec`/`shutil.which` with the shared probe — one source of truth for "what the gate runs" and "what doctor installs". **`security` (full)** adds `pip-audit`.
- **`duplicate-strings`** declares the `node` runtime (optional; global `find-duplicate-strings` binary as an `alternative`) — the real external dep it previously only discovered at subprocess time.

All optional (both gates degrade gracefully today), so **behavior-preserving**. Full suite green (3330); new tests cover the probe kinds, name/import-name split, the pip-installed-binary case, env probing, and each gate's declarations.

## Next
Remaining gates by category; wire `requirement_block_result` into required-tool gates' `run()`; registry enforcement test (lands last); then `sm doctor --required-deps` + the v2 Action.

Implements a requirements contract for external tool “presence” by extending `Requirement` with `probe` (binary|import|env|none) and `import_name`, and adding `BaseCheck._module_available()` plus `BaseCheck.is_requirement_satisfied()`/probe-aware `missing_requirements()`. This replaces scattered PATH/import/spec checks with a single, consistent probing model and ensures requirements are serialized into the manifest with the new fields.

Updates the `security` and `duplicate-strings` gates to declare dependencies as optional requirements and to resolve/invoke the same executables/modules they probe: `semgrep` and `node` are run via venv-aware `find_tool()`/resolved paths, `detect-secrets` uses the correct `import_name` (`detect_secrets`), and `pip-audit` is probed as an import (`pip_audit`) since it runs as `python -m pip_audit`. Also fixes probe behavior drift (notably `semgrep` and `node` invocation) and aligns env probing to honor `alternatives`.

Adds unit tests covering probe kinds, `import_name` handling, and the specific “resolved path is what gets invoked” cases; test suite passes (3330 tests).

### DIFF
--- a/.willville.json
+++ b/.willville.json
@@ -1,7 +1,7 @@
 {
   "agent": {
-    "status": "Made `sm sail` drive a change all the way to a green PR on its own, so there's one command to run instead of a handful — PR #302, green",
-    "direction": "Point everyone at that single command and tuck the rest of the tooling behind it",
-    "last_update": "2026-06-19T22:01:00Z"
+    "status": "Teaching each check to declare what tools it needs, so one place can install them — did the two tricky ones (security scanners, duplicate-strings) on PR #306, stacked on #305",
+    "direction": "Convert the rest of the checks the same way, then let doctor read the list and have the GitHub Action install from it",
+    "last_update": "2026-06-20T02:25:00Z"
   }
 }
--- a/slopmop/checks/base.py
+++ b/slopmop/checks/base.py
@@ -923,7 +923,9 @@ class BaseCheck(ABC):
             names += [alt.replace("-", "_") for alt in req.alternatives]
             return any(_module_available(name) for name in names)
         if probe == "env":
-            return bool(os.environ.get(req.name))
+            # Honour alternatives like binary/import do (e.g. GH_TOKEN with a
+            # GITHUB_TOKEN alternative) — any one set satisfies it.
+            return any(os.environ.get(n) for n in (req.name, *req.alternatives))
         return True  # "none" — defer to the installing layer
 
     def missing_requirements(self, project_root: str) -> List[Requirement]:
--- a/slopmop/checks/base.py
+++ b/slopmop/checks/base.py
@@ -261,6 +261,22 @@ def find_tool(name: str, project_root: str) -> Optional[str]:
     return shutil.which(name)
 
 
+def _module_available(import_name: str) -> bool:
+    """Return True if a Python module is importable, without importing it.
+
+    Uses ``importlib.util.find_spec`` rather than ``import_module``: importing
+    some packages (e.g. bandit) pulls in stevedore, which enumerates every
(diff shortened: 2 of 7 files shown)


ScienceIsNeato:
[fixed_in_code] Fixed: _get_tool_path now resolves the global find-duplicate-strings binary via find_tool(project_root), and _build_command resolves the node runtime via find_tool too (threaded from run()) instead of invoking bare 'node'. Both now use the same primitive the requirements contract probes, so declaration and behavior can't drift. Added a regression test asserting the resolved node path is used.

ScienceIsNeato:
[fixed_in_code] Fixed: _run_semgrep now resolves semgrep via resolve_requirement_path (the same find_tool the availability probe uses) and invokes that path, so a venv-only semgrep is run correctly instead of failing on a bare name. Regression test asserts the resolved path is used.

ScienceIsNeato:
[fixed_in_code] Fixed (my modeling error): pip-audit is now probe='import', import_name='pip_audit', matching its  invocation — instead of probe='binary' which looked for a pip-audit script that may not exist. Detection now matches invocation. Test asserts the probe model.

ScienceIsNeato:
[fixed_in_code] Fixed: SecurityCheck.requirements() now declares the fixed scanner set its run() actually executes (bandit, semgrep, detect-secrets, pip-audit), independent of the scanners config — so doctor/the Action can't under-install relative to runtime. Test asserts the full set is declared even when config trims it.

ScienceIsNeato:
[fixed_in_code] Fixed: the env probe now checks req.name plus alternatives (any set satisfies it), consistent with binary/import probes. Added a test for GH_TOKEN-with-alternative.

PR #310, opened 2026-06-20
refactor(doctor): unify on requirements() — retire required_tools + REQUIRED_TOOLS

ScienceIsNeato:
The full unification (your call): `requirements()` becomes the **single source** of "what external tools do gates need, and how does a user install them." Both legacy sources and the three legacy class attrs are retired.

## The reconciliation
The blocker was that two sources served different audiences:
- `requirements()` → downstream **Action** (install `bandit==1.7.5`).
- `REQUIRED_TOOLS` (`detection.py`) → slop-mop's **own env-doctor** (`pipx install slopmop[security]`).

Resolved by giving `Requirement` an **`install_hint`** — the user-facing remediation command — set via `pip_cli_requirement(..., extra="security")` (→ `pipx install slopmop[security]`) or an explicit SDK URL. One Requirement now carries both the per-tool pin (for the Action) and the extras-group hint (for the env-doctor).

## Consumers migrated
- **`detection.py`**: the hardcoded `REQUIRED_TOOLS` list is gone; `_detect_tools()` derives `(tool, gate, install_hint)` from the new `tool_inventory.gate_tool_inventory()`.
- **`sm_env.py`** (both loops + install-hint grouping), **`gate_preflight.py`**, **`config.py`**, **`doctor.py`** → all read `requirements()` / `missing_requirements()` / `is_requirement_satisfied()`.

## Retired
`BaseCheck.required_tools` / `required_tool_versions` / `install_hint` class attrs, and `required_tools=[...]` / `install_hint="path"` on all ~11 gates.

## Why this is strictly better
The derived inventory now covers **ruff, node, actionlint** too — tools the hand-maintained `REQUIRED_TOOLS` had silently missed. Enforcement tests assert the inventory stays complete and every row carries an install hint.

Full suite green (3339; the 2 `test_refit_drain_functional` failures are pre-existing — fail on clean main, formatter-version flakiness).

## Next (the original goal this unlocks)
`sm doctor --required-deps` manifest emitter (additive, now trivial — one source) → the **v2 Action** consumer + the org-wide manual-scour audit.

## Summary

This PR makes `requirements()` the single source of truth for both gate external-tool needs and the user install/remediation commands, removing the prior split between per-tool pins, legacy per-gate class attributes, and a hardcoded `REQUIRED_TOOLS` list.

**Core changes**
- Added `Requirement.install_hint` plus `Requirement.resolved_install_hint()` to produce environment-doctor install guidance (uses `install_hint` when provided; otherwise falls back by requirement kind).
- Extended `pip_cli_requirement(..., extra=...)` to auto-generate install hints for extras groups (e.g., `pipx install slopmop[security]`).
- Added `checks/tool_inventory.py: gate_tool_inventory()` to derive a deduped `(tool, gate, install_hint)` registry from all gates’ `requirements()` (skips `env`-kind requirements), replacing the hardcoded tool inventory.
- Updated gates to declare dependencies via `requirements()` only (retired legacy `required_tools`, `required_tool_versions`, and gate-level `install_hint`).

**Behavioral fixes / migrations**
- Doctor readiness + preflight now treat missing **optional** tools as WARN (only non-optional missing tools block).
- Gate checks now instantiate using each repo’s `.sb_config.json`, so `requirements()` reflects repo configuration.
- Version checks now interpret pinned requirements as **minimum floors** (warn when installed versions are older, rather than enforcing exact pins).
- Missing-tool remediation messages are sourced from the unified inventory/`resolved_install_hint()`, including env-kind-specific variable hints.

**Consumer updates**
- `cli/detection.py`, `doctor/sm_env.py`, `doctor/gate_preflight.py`, `cli/config.py`, and `cli/doctor.py` migrated to the unified requirements/tool-inventory flow.

**Testing**
- Updated and extended unit tests to mock `gate_tool_inventory()`, validate deduping/version reporting, and cover `resolved_install_hint()` + env skipping + migration behavior.
- Test suite passes (3339 passing tests).

### DIFF
--- a/slopmop/checks/base.py
+++ b/slopmop/checks/base.py
@@ -624,13 +624,15 @@ class Requirement:
         }
 
     def resolved_install_hint(self) -> str:
-        """The install command to show a user, with a kind-based fallback."""
+        """The remediation to show a user, with a kind-based fallback."""
         if self.install_hint:
             return self.install_hint
         if self.kind == "python":
             return f"pip install {self.name}"
         if self.kind == "npm":
             return f"npm install -g {self.name}"
+        if self.kind == "env":
+            return f"Set the {self.name} environment variable"
         return f"Install {self.name}"
 
 
--- a/slopmop/checks/base.py
+++ b/slopmop/checks/base.py
@@ -602,6 +602,12 @@ class Requirement:
     alternatives: tuple[str, ...] = ()
     probe: str = ""  # "" = default by kind; "binary" | "import" | "env" | "none"
     import_name: str = ""  # for probe="import" when it differs from name
+    # User-facing install command for slop-mop's OWN env-doctor remediation —
+    # e.g. "pipx install slopmop[security]" or "Install Flutter SDK: …". Distinct
+    # from how the downstream Action installs (name + version + kind): this tells
+    # a slop-mop *user* how to fix their environment. Empty ⇒ doctor falls back
+    # to a kind-based default ("pip install <name>").
+    install_hint: str = ""
 
     def to_manifest(self) -> Dict[str, Any]:
         """Serialize to the deterministic manifest shape doctor/the Action read."""
@@ -613,8 +619,19 @@ class Requirement:
             "alternatives": sorted(self.alternatives),
             "probe": self.probe,
             "import_name": self.import_name,
+            "install_hint": self.install_hint,
             "reason": self.reason,
         }
+
+    def resolved_install_hint(self) -> str:
+        """The install command to show a user, with a kind-based fallback."""
+        if self.install_hint:
+            return self.install_hint
+        if self.kind == "python":
+            return f"pip install {self.name}"
+        if self.kind == "npm":
+            return f"npm install -g {self.name}"
+        return f"Install {self.name}"
 
 
 @dataclass(frozen=True)
@@ -642,7 +659,12 @@ def build_requirements_document(requirements: "Requirements") -> Dict[str, Any]:
 
 
 def pip_cli_requirement(
-    name: str, version: Optional[str], reason: str, *, optional: bool = False
+    name: str,
+    version: Optional[str],
+    reason: str,
+    *,
+    optional: bool = False,
+    extra: Optional[str] = None,
 ) -> Requirement:
     """A pip-installed tool invoked as a CLI binary — the common gate case.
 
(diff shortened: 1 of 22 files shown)


ScienceIsNeato:
[fixed_in_code] Fixed in f168b6c: _missing_required_tools counts only non-optional missing tools, so an absent optional tool degrades the gate (WARN), not blocked.

ScienceIsNeato:
[fixed_in_code] Fixed in f168b6c: GateReadinessCheck filters to required (non-optional) missing requirements before marking a gate blocked.

ScienceIsNeato:
[fixed_in_code] Fixed in f168b6c: _collect_version_violations reads each requirement's pin as a floor (warn if older) instead of the removed required_tool_versions attr.

ScienceIsNeato:
[fixed_in_code] Fixed in f168b6c: resolved_install_hint() returns 'Set the {name} environment variable' for env-kind requirements.

ScienceIsNeato:
[fixed_in_code] Fixed in f168b6c: GateReadinessCheck instantiates each gate with the repo's .sb_config.json (via _load_repo_config) so config-dependent requirements() reflect the repo.

ScienceIsNeato:
[fixed_in_code] Fixed in f168b6c: sm doctor --gates renders requirements() using the repo's loaded config, not {}.

ScienceIsNeato:
[fixed_in_code] Fixed in f168b6c: both GateReadinessCheck and sm doctor --gates load and pass the repo config to get_check.

ScienceIsNeato:
[fixed_in_code] Fixed in f168b6c: the last reader of a retired attr (required_tool_versions, in _collect_version_violations) is rewired to requirements(). Nothing reads required_tools/required_tool_versions/install_hint now.

ScienceIsNeato:
[fixed_in_code] Fixed in f168b6c: the enable-path test now asserts cmd_config(args) == 0.

ScienceIsNeato:
[invalid_with_explanation] The disagreement is theoretical for the current tool set. ToolInventoryCheck uses find_tool (binary presence) as the install-prompt heuristic; GateReadinessCheck uses the probe-aware missing_requirements() as the authoritative per-gate signal. Every currently-declared tool ships a console script — bandit/semgrep/detect-secrets/pip-audit all install a CLI via pip, and dart/flutter are binaries — so find_tool and the import/binary probes agree in practice. If a module-only tool (no console script) is ever declared, we'd route ToolInventoryCheck through the probe too; until then this is a non-issue and unifying it now adds the Requirement plumbing to the env-doctor loop for no behavioral gain.

ScienceIsNeato:
[fixed_in_code] Fixed in 34423bb: gate_tool_inventory() now takes a config param, and detection._detect_tools + sm_env's ToolInventoryCheck/GateReadinessCheck load and pass the repo's .sb_config.json — so config-gated requirements (run_actionlint off, reduced scanners) are honoured, matching the readiness checks. Added a test asserting the config is forwarded to get_check.

ScienceIsNeato:
[fixed_in_code] Fixed in 34423bb: the all-ignoring monkeypatch lambda is now 'lambda *_a: None' (Ruff ARG005).

Issue #11, opened 2026-02-06
feat: Add quality:dead-code gate wrapping vulture


ScienceIsNeato:
## Summary

Add a `quality:dead-code` gate that wraps [vulture](https://pypi.org/project/vulture/) to detect unused functions, classes, imports, variables, and unreachable code via static AST analysis.

## Motivation

Dead code accumulates silently — unused functions after refactors, orphaned imports, unreachable branches after control flow changes. It increases cognitive load, bloats test coverage denominators, and creates false confidence that code paths are maintained. Catching it automatically in the commit/PR loop prevents rot.

### Tool
- **vulture** v2.14+ (MIT, Python 3.13 support, canonical dead-code finder)
- Pattern: Tool-wrapper (same as `complexity.py` wraps `radon`)
- File: `slopmop/checks/quality/dead_code.py`

### How It Works
1. Build command: `python -m vulture <src_dirs> --min-confidence <n> --exclude <patterns>`
2. Run via `_run_command()` with timeout
3. Parse output lines: `file.py:42: unused function 'foo' (80% confidence)`
4. Exit code 3 = dead code found, 0 = clean
5. Handle tool-not-installed gracefully (returncode 127 → skip with message)

### Config Schema
| Field | Type | Default | Description |
|---|---|---|---|
| `min_confidence` | integer | 80 | Minimum confidence to report (60-100) |
| `exclude_patterns` | string[] | `["**/venv/**", "**/.venv/**", "**/node_modules/**", "**/test_*"]` | Glob patterns to exclude |
| `src_dirs` | string[] | `["."]` | Directories to scan |
| `whitelist_file` | string | `""` | Path to vulture whitelist file |

### Registration
- Register in `_register_crosscutting_checks()` in `slopmop/checks/__init__.py`
- Export from `slopmop/checks/quality/__init__.py`
- Add to `commit` and `pr` profiles

### Dependencies
- Add `vulture>=2.14` to `requirements.txt`

### Tests
- Test file: `tests/unit/test_dead_code_check.py`
- Cover: config defaults, output parsing, tool-not-installed handling, exclude patterns, whitelist support

## Estimated Effort
~120 lines check + ~150 lines tests (~1 hour)

## Acceptance Criteria
- [ ] `sm validate quality:dead-code` runs vulture and reports findings
- [ ] Graceful skip when vulture not installed
- [ ] Configurable min_confidence, exclude patterns, src_dirs, whitelist
[...]

PR #326, opened 2026-07-05
fix: tech-debt phase 2 — unpinned-ref check, venv unification, registry smoke, drain-test root cause

ScienceIsNeato:
Phase 2 of the tech-debt audit — the slop-mop-side items (the action-repo `project-install` input ships separately).

## 1. `github-actions-hygiene` now catches unpinned action refs (audit #5)
New `unpinned-action-ref` check: any `owner/repo@ref` that isn't a **full 40-hex commit SHA** is a finding — mutable tags/branches can be repointed to different code, and short SHAs are forgeable. `./local` and `docker://` refs exempt. Deliberate moving tags opt out via the new `allow_unpinned` config (prefix-matched); slop-mop allowlists its own `slop-mop-action@v2` dogfood ref in pyproject. Closes the gap where CodeRabbit caught unpinned actions our own gate missed. 6 new tests; existing fixtures SHA-pinned; the deprecated-major test now asserts both findings coexist on a deprecated tag.

**Consumer note:** repos with tag-pinned actions will see new findings on upgrade — that's the check working; remediation is pin-or-allowlist.

## 2. Venv detection unified (audit #7)
`type_checking._detect_venv_path` (private duplicate) → `mixins.detect_venv_path`, sharing the new `VENV_DIR_NAMES` constant with `has_project_venv`. One source of truth for "which dirs count as the project venv, in what order" — the exact divergence class behind this week's environment bugs. No behavior change.

## 3. Real-registry smoke tests (audit #9)
`test_registry_smoke.py` exercises genuine registration end-to-end: every gate registers + instantiates with `full_name == registered name`, remediation priorities resolve, the remediation sort is total/stable, and the hand-curated order references only real gates. The refit suite's `_FakeRegistry` can no longer hide a real registration regression.

## 4. Drain functional tests: root cause found ≠ the theory (audit #11)
The two `test_refit_drain_functional` tests that failed locally all week were *not* victims of machine git hooks (first theory, instrumented and falsified): **drain resolves formatters via the process-global registry, which is empty when the file runs in isolation** — drain silently no-ops, so no style commit. They passed in CI only because earlier test files happened to register gates. The autouse fixture now registers checks and *also* isolates git config (`GIT_CONFIG_GLOBAL`/`GIT_CONFIG_SYSTEM` → CI-faithful bare environment, defense-in-depth). Same order-dependence fixed in `test_name_and_registration`.

## Proof
**3,372 passed with zero deselects** — first fully-clean local run of the suite including the two previously-quarantined tests. Hygiene gate passes on slop-mop's own workflows (allowlist verified).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -186,3 +186,9 @@ pip_audit_ignore_vulns = [
 pip_audit_ignore_vulns = [
     "GHSA-5239-wwwm-4pmq",  # pygments ReDoS (CVE-2026-4539), low severity, no fix available
 ]
+
+[tool.slopmop.myopia.gates."github-actions-hygiene"]
+# The dogfood deliberately tracks the moving @v2 major tag — that tag is the
+# delivery channel for action fixes and validating what consumers resolve is
+# the dogfood's whole point. Everything else must be SHA-pinned.
+allow_unpinned = ["scienceisneato/slop-mop-action"]
--- a/slopmop/checks/mixins.py
+++ b/slopmop/checks/mixins.py
@@ -73,6 +73,14 @@ PYTHON_SOURCE_PATH = "system_path"
 PYTHON_SOURCE_NOT_FOUND = "not_found"
 
 
+# Recognized project-venv directory names, in priority order. The single
+# source of truth for every consumer that resolves a project environment —
+# detect_venv_path (pyright config), has_project_venv, and the mixin's
+# python resolution must agree on this, or gates silently analyze
+# different environments.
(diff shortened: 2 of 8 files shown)


ScienceIsNeato:
[fixed_in_code] Fixed: _FULL_SHA_RE now matches [0-9a-fA-F]{40} (case-insensitive hex), and test_uppercase_full_sha_pin_is_clean covers an uppercase 40-hex ref as a regression guard.

PR #327, opened 2026-07-05
refactor: tech-debt phase 3 — one gate-name parser, one enablement truth, named timeouts

ScienceIsNeato:
Phase 3 of the tech-debt audit: the config-unification arc that retires the divergence bug class, plus the agreed dogfood lead-in.

## 1. `GateRef` — one parser for `category:gate` (audit #4)
New `core/gate_config.py` hosts the single parser for the stringly-typed gate-name format. All **25+ ad-hoc `split(":", 1)` sites** across cli/core/doctor/reporting/migrations now go through it; a repo-wide grep confirms **zero remain**. Any future change to the naming scheme is one function.

## 2. One enablement truth — and it fixed another live divergence (audit #4)
`gate_enablement()` / `is_gate_enabled()` carry the executor's semantics (the implementation that decides what actually runs): `disabled_gates` → category-level `enabled: false` → gate-level explicit `false` → enabled. Executor, doctor preflight, and `sm config` all **delegate** now.

**The fix this surfaced:** preflight and `sm config` previously ignored a category-level `{"<category>": {"enabled": false}}` — reporting gates as enabled that the executor would skip. Same bug family as #310 and the Phase-1 preflight bug; regression tests pin both consumers.

**The deletion this justified:** `SlopmopConfig.is_gate_enabled` / `CategoryConfig.is_gate_enabled` had **zero production consumers** and defaulted to `enabled=False` — the *opposite* of the executor. Dead API with wrong semantics got deleted, not reconciled; a tombstone test blocks a rival API from reappearing.

## 3. Named timeout tiers (audit #10)
`checks/timeouts.py` defines six intent-named tiers (`PROBE_TIMEOUT` 5s → `EXHAUSTIVE_TASK_TIMEOUT` 900s). **61 literals across 28 gate files** migrated with **zero value changes** (exact-match only; the stray 10/15/180s literals are left for opportunistic normalization, documented in the module).

## 4. Dogfood exercises `project-install`
`action-dogfood.yml` now provisions the project venv through the action's own `project-install` input instead of a hand-rolled workflow step — so the v2 input added in action #6 is tested on every PR.

## Poetic footnote
Mid-commit, the `ambiguity-mines` gate blocked me for defining a second `_as_dict` with different semantics than an existing one — the tooling enforcing this PR's own thesis. Renamed to `_mapping_or_empty`.

Full suite: **3,386 green** (+14 gate_config tests), swab clean.

Unified gate parsing and enablement semantics, and standardized subprocess timeouts across the checks tree.

- Added `core/gate_config.py` with `GateRef`, `gate_enablement()`, and `is_gate_enabled()`.
- Replaced ad hoc `category:gate` splitting across core, CLI, doctor, migrations, reporting, and registry code.
- Removed legacy `is_gate_enabled` methods from config classes and added regression tests.
- Centralized timeout literals into `checks/timeouts.py` and migrated checks to the new named tiers without changing values.
- Updated dogfood workflow to use the action’s `project-install` v2 input.

### DIFF
--- a/.github/workflows/action-dogfood.yml
+++ b/.github/workflows/action-dogfood.yml
@@ -47,23 +47,6 @@ jobs:
           # from the real gate and inflates "provisional".
           fetch-depth: 0
 
-      # Type-aware gates (pyright/mypy) resolve imports against a detected venv
-      # (./venv, ./.venv, or $VIRTUAL_ENV). The v2 action installs slop-mop and
-      # the gate tools, but NOT this project's own dependencies — so without a
-      # project venv, pyright can't resolve our imports (e.g. the tomli fallback,
-      # which ships transitively via semgrep) and type-blindness fails on
-      # phantom "unknown type" findings. A real consumer installs their deps
-      # before this step; we mirror that here.
-      - name: Set up Python
-        uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405  # v6.2.0
-        with:
-          python-version: "3.12"
-      - name: Install project deps into ./venv (for type-aware gate resolution)
-        run: |
-          python -m venv venv
-          ./venv/bin/pip install --upgrade pip
-          ./venv/bin/pip install -e ".[all]"
-
       # actionlint is a system tool the manifest can't install (Go binary). It's
       # applicable here (we have workflows), so provide it on PATH — otherwise
       # myopia:github-actions-hygiene warns-and-skips its actionlint pass. A real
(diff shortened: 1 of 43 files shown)


ScienceIsNeato:
[fixed_in_code] Fixed in d73cfa0: the dart-format timeout message now interpolates SLOW_TOOL_TIMEOUT ('timed out after {SLOW_TOOL_TIMEOUT}s') instead of hardcoding '2 minutes' — no drift if the tier changes.

ScienceIsNeato:
[fixed_in_code] Fixed in d73cfa0: same treatment — flutter-tests timeout message interpolates EXHAUSTIVE_TASK_TIMEOUT instead of '15 minutes'.

ScienceIsNeato:
[fixed_in_code] Fixed in d73cfa0: gate_config now imports Mapping from collections.abc, uses builtin dict[str, Any]/tuple[bool, str] generics, and drops the quoted GateRef return annotation (the module already has from __future__ import annotations).

ScienceIsNeato:
[invalid_with_explanation] The Python pin didn't leave — it moved inside the action. project-install executes as a step INSIDE the slop-mop-action composite, after the action's own actions/setup-python step (its python-version input, default '3.12'). setup-python's GITHUB_PATH export applies to all subsequent steps in the job, including composite steps, so 'python -m venv' in project-install resolves to the action's pinned 3.12 — not whatever ubuntu-latest ships. The removed workflow-level setup-python only ever fed the manual venv step this PR deleted. Verified in this PR's dogfood run: the venv was created and used, hull rating A+.

PR #331, opened 2026-08-10
fix: six onboarding barnacles found running slop-mop cold on a stranger's repo

ScienceIsNeato:
We onboarded **`botingw/rulebook-ai`** (MIT, ~8k LOC Python, 603★, *already* running ruff + mypy + pytest + CI) as a case-study subject. The run never reached a plan — it hit six real product problems first. This PR fixes all six and keeps the evidence.

Every fix was verified against the live repro, not just unit tests.

## The two that shaped the first-run experience

**#1 — Vendored copies multiplied findings 5–7×.** rulebook-ai distributes tool-starter templates, so `tool_starters/llm_api.py` exists as **7 byte-identical copies** (verified by md5). Every gate reported the same defect once per copy: 1 unused import → 5 findings, 1 oversized function → 7, 1 placeholder string → 7 "potential secrets". **~57 reported for ~20 real problems** — landing squarely on the impression that decides whether someone keeps using the tool.

`checks/duplicate_files.py` collapses findings whose files are byte-identical (size + sha256) *and* that share a line and message. Survivor keeps the shortest path and gains `[also in N identical copies of this file]`. Hooked in the executor, so one place fixes every gate. **Deliberately conservative:** a file that diverged by one byte, two different messages on the same line, and unreadable/absent/huge files are never merged — hiding a genuinely divergent bug would be worse than noise.

**#2 — The grade never moved.** The letter counts failing *gates*, so a legacy repo sits at F (4+ failing) through an enormous amount of real work. Our own pass cleared 3 gates and 65% of findings with the grade unmoved — which reads as "nothing happened" to exactly the inherited-codebase user `sm refit` exists to serve.

`HullGrade` now carries `findings` + `previous_findings`/`findings_delta`, rendering as `F — scuttled · 20 findings (down 37)`. **The letter itself is untouched** — the Action's `minimum-grade` input depends on that contract, so the new numbers ride alongside rather than shifting any boundary. v3 swab/scour schemas extended to match.

## The four that blocked or misled

**#3 — Refit hard-blocked by patch drift.** `black 26.5.0` vs a `26.5.1` pin made `sm refit --start` refuse to run. A user one patch behind couldn't onboard at all. Version drift now **WARNs**; missing tools and validator rejects still **FAIL** (they genuinely can't run).

**#4 — Refit demanded you justify init's own choices.** `sm init` disables gates by its own detection, then `refit --start` listed those as pending and refused to plan until each was approved *or given a bug reference*. init now stamps `disabled_by: "init"|"user"`, and refit auto-accepts the tool-owned ones. Unknown provenance (hand-edited configs) still asks — the safe direction.

**#5 — A gate failed with zero findings.** `missing-annotations` reported `FAILED: 0 type error(s) found` with empty output whenever mypy exited non-zero for a *non-type* reason — here, duplicate module names caused by #1's vendored copies. Now a could-not-run **ERROR** carrying mypy's own words (`Duplicate module named "llm_api"`) and its suggested resolutions. Genuine type errors still FAIL.

**#6 — Inconsistent remediation guidance.** `ambiguity-mines` supports `exclude_dirs` but never said so, unlike its sibling `repeated-code`. Vendored copies should be excluded, not refactored; the hint now says that and prints the command.

## Evidence kept
`DOCS/case-studies/` holds the baseline/final scour artifacts and the full barnacle write-up with the refit progress log.

## Notes
- `#4`'s fix is narrow by design: refit still asks you to review each **enabled, runnable** gate. That's its core fidelity design, not the barnacle — the barnacle was justifying the tool's *own* disables.
- Suite **3,409** green. Along the way slop-mop blocked my own commits five times (code-sprawl, four type-annotation gaps) — every one a real defect in the patch.

## Summary

- Collapse findings from byte-identical files.
- Add finding counts and deltas without changing hull grades.
- Change tool version drift from `FAIL` to `WARN`.
- Track `sm init` gate-disable provenance for `refit`.
- Report non-type mypy failures as errors with diagnostic output.
- Improve `ambiguity-mines` guidance for `exclude_dirs`.
- Add case-study documentation, schemas, and focused tests.


ScienceIsNeato:
[fixed_in_code] Fixed in bde3363: the group key is now (content_hash, line, column, rule_id, message) — two diagnostics sharing a line and message text but differing in column or rule are kept separate. Tests added for both distinctions.

ScienceIsNeato:
[fixed_in_code] Fixed in bde3363: the merged Finding carries its group's earliest original position, so survivors stay in place instead of sinking to the end. test_survivor_keeps_its_original_position covers it.

ScienceIsNeato:
[fixed_in_code] Fixed in bde3363 (same change as the ordering thread): ordering is now computed from the earliest position within each group rather than id() of the new object.

ScienceIsNeato:
[fixed_in_code] Fixed in bde3363: paths are resolved and checked with relative_to(project_root); anything escaping the root is never hashed. test_paths_escaping_project_root_are_never_hashed covers both the '..' and absolute-path cases.

ScienceIsNeato:
[fixed_in_code] Fixed in bde3363: the cache-hit return now passes through the same collapse, so caches written before this existed no longer surface duplicates or inflated counts. The operation is idempotent.

ScienceIsNeato:
[fixed_in_code] Fixed in bde3363: a non-green gate with no structured findings now counts as 1 rather than 0, so a failing run can never render 'F — scuttled · 0 findings'. Test added.

ScienceIsNeato:
[fixed_in_code] Fixed in bde3363: _previous_findings_count rejects bools (bool is an int subclass) and negatives, returning None so no delta is shown. Covered for -1, True, a string, and null.

ScienceIsNeato:
[fixed_in_code] Fixed in bde3363: compute_hull_grade applies the same validation at the constructor boundary — bool or negative previous_findings becomes None. Tests added for both.

ScienceIsNeato:
[fixed_in_code] Fixed in bde3363: _stamp_auto_disabled_provenance now reads the existing on-disk config and preserves a prior 'user' marker, so a hand-disabled gate stays user-owned across sm init re-runs. Three tests cover preserve-user, keep-init, and first-run-with-no-existing-config.

ScienceIsNeato:
[fixed_in_code] Fixed in bde3363: dropped the unused tmp_path fixtures from the two tests that never referenced it.

ScienceIsNeato:
[fixed_in_code] Fixed in bde3363: tagged the fenced block as `text` (MD040).

PR #332, opened 2026-08-10
fix(gates): four gate defects found running a refit on a stranger's repo

ScienceIsNeato:
Four gate defects surfaced by running a real refit against
[botingw/rulebook-ai](https://github.com/botingw/rulebook-ai) — a stranger's
repo, driven from a failing baseline to A+ as a case study.

Each one cost real time during that run, which is how they were found.

### `string-duplication` reported paths relative to nothing
`_format_findings` called `os.path.relpath(file_path)` with no start dir, so
paths resolved against the process cwd. Underneath it, the scanner ran in a
tempdir: macOS `tempfile` returns `/var/folders/...` while the scanner
reports the resolved `/private/var/folders/...`, so replacing only the
unresolved spelling matched the suffix and left `/private` glued to the
front of every path. Findings pointed at files that did not exist.

### `dangling-references` could not tell code from prose
Python subscript-then-call is byte-identical to a markdown link, so a
documented `handlers[parsed_args.command](parsed_args)` inside a fenced
python block was reported as a broken link to a target named `parsed_args`.
On rulebook-ai this was the *only* remaining "broken link" after 12 genuine
ones were fixed — the last thing between the gate and green was not a defect.

Fenced blocks are now skipped and inline code spans blanked (preserving line
length, so `` [`spec.md`](../spec.md) `` still gets its target checked).

### A security scanner that never ran was reported as a finding
`_scanner_failed_to_start()` already existed, with a comment naming this
exact failure mode — *"Reporting 'No module named detect_secrets' as SLOP
DETECTED tells a user they have a leaked secret when they have a broken
install."* It was never called from anywhere.

On a repo whose venv lacked pip-audit, `dependency-risk` failed with "1
security scanner(s) found issues" and a finding of "pip-audit found issues":
naming a vulnerability that does not exist, while hiding that nothing had
been audited. bandit, semgrep and pip-audit now route startup failures to a
warned result. A scanner that ran and genuinely errored still fails.

### `refit --iterate` named the wrong failing gate
A targeted scour runs the gate *plus its dependencies*, so when a dependency
fails the run stops on a different gate than the one being iterated — but the
message named the iterated gate and pointed at its log, which had not been
rewritten. Iterating `missing-annotations` while `sloppy-formatting` failed
reported "stopped on failing gate: missing-annotations" beside a stale
13-error mypy log. The artifact's `first_to_fix` was right all along.

---

Also updates the case-study docs with the final result: **A+ — shipshape, 0
findings, 21 gates passing**, from a baseline of 8 failing gates. rulebook-ai's
own suite went 53 passing / 3 failing → 81 passing / 0 failing.

- Remap duplicate-string paths to `project_root`, with current-directory fallback.
- Ignore code blocks and inline code in dangling-reference checks.
- Report unavailable security scanners as degraded warnings.
- Identify the actual failing gate during refit iterations.
- Add regression tests and document the fixes.


ScienceIsNeato:
[fixed_in_code] Fixed in 7934ec7. Assertion now uses os.path.join("pkg", "mod.py") so it holds on Windows, and both fixtures gained fileCount to match real find-duplicate-strings output.

ScienceIsNeato:
[fixed_in_code] Fixed in 7934ec7. The inline-code regex is now delimiter-aware, so a run of N backticks closes only on a run of exactly N and shorter runs inside the span no longer terminate it early. Regression test: test_multi_backtick_span_closes_on_equal_run.

ScienceIsNeato:
[fixed_in_code] Fixed in 7934ec7. Code spans are now matched across the whole document rather than per line, so a span containing a line ending is blanked on both lines. Newlines are preserved during blanking so findings keep their real line numbers. Per CommonMark a span cannot contain a blank line, and an unmatched backtick stays literal instead of swallowing the rest of the file. Regression tests: test_code_span_wrapping_lines_is_blanked, test_unclosed_backtick_does_not_swallow_later_links.

ScienceIsNeato:
[fixed_in_code] Good catch, and the most important one here — this was strictly worse than the bug it replaced, since a broken install would have rendered as 'All security checks passed'. Fixed in 7934ec7: SecurityLocalCheck.run now collects warned sub-results and returns a WARNED result when nothing failed outright. _warnings_only_result and _collect_findings moved up from SecurityCheck so both share them, and a did_not_run flag keeps 'never started' guidance (install it) distinct from 'ran, no patch available' guidance. Regression tests: test_local_run_warns_when_scanner_never_started plus test_local_run_still_passes_when_scanners_are_clean to pin the clean path.

ScienceIsNeato:
[fixed_in_code] Fixed in 7934ec7. When actual_gate differs from the iterated gate and first_to_fix carries no log_file, no log line is printed at all — showing the requested gate's stale log was the original complaint this commit set out to fix.

ScienceIsNeato:
[fixed_in_code] Fixed in 7934ec7. Added 'assert result.findings[0].line == 5' — the blanking preserves newlines specifically so positions survive, and without this assertion a renumbering regression would pass silently.

ScienceIsNeato:
[fixed_in_code] Fixed in 7934ec7. The test now imports _remap_scan_paths and calls it with stdout, tmp_dir and project_root instead of reimplementing the loop — as written it would have passed even with the helper broken, which defeats the point of a regression test.

ScienceIsNeato:
[fixed_in_code] Fixed in 7934ec7 — right about the residual risk. _format_findings now realpaths both the file path and the start path before relpath, so a symlinked project root cannot reintroduce the /Users vs /private/Users escape. Regression test: TestSymlinkedProjectRoot.test_symlinked_root_still_yields_relative_paths, which builds a real symlinked root and feeds in the resolved path the scanner actually emits.

ScienceIsNeato:
[invalid_with_explanation] Declining this one. slop-mop's own ruff config does not enable the UP ruleset, so UP035 is coming from CodeRabbit's ruff rather than the project's — and typing.List/Tuple is the convention across 94 files in slopmop/. Converting one file would leave it inconsistent with the other 93 without fixing anything. Worth doing as a repo-wide sweep if the project wants to adopt builtin generics, but that is its own change, not this PR.

PR #333, opened 2026-08-11
docs(case-study): correct the rulebook-ai test baseline

ScienceIsNeato:
The rulebook-ai progress log claimed their suite went "53 passing / 3 failing → 81 passing". That reads as though we inherited three broken tests and fixed them. We did not.

At the merge-base (`7bc8531`) all **56** of their tests pass. The three failures were our own regression — the URL-scheme validation added for bandit B310 removed `file:` support, which their community-index integration tests rely on. The `git stash` check that appeared to prove the failures were pre-existing only removes *uncommitted* work, so the already-committed breakage survived it and every "baseline" run reproduced it.

Corrected to **56 passing at the merge-base → 81 at the end** (25 added), with the reason stated inline. Leaving the old number would have credited us with a repair we never made, in the document that argues for reporting our own defects.

Surfaced while writing the number up for the website case study.

Updates the rulebook-ai case study to correct the test baseline: 56 tests passed at merge-base and 81 passed at completion. Clarifies that the three earlier failures came from a project regression removing `file:` URL support.

### DIFF
--- a/.willville.json
+++ b/.willville.json
@@ -1,7 +1,7 @@
 {
   "agent": {
-    "status": "Ran slop-mop cold on a stranger's repo (rulebook-ai) as a case study and drove it from failing to a clean bill of health — the run shook out four of our own bugs, up on PR #332",
-    "direction": "Write the case study up for the site now that we have the before and after numbers",
-    "last_update": "2026-08-10T21:45:00Z"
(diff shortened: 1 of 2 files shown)


ScienceIsNeato:
[fixed_in_code] Both fixed. The paragraph is now hard-wrapped to match the rest of the document. On the numbering: I dropped the reference rather than renumbering it — this regression was not a barnacle at all. Barnacles are friction in slop-mop; this was our bug in *their* code. It now points at step 10 of the progress log, which is where it was actually found and fixed.

PR #334, opened 2026-08-11
docs: add the 2.13.0 changelog section

ScienceIsNeato:
Prep for cutting 2.13.0. The release workflow reads `CHANGELOG.md` for the release body and **fails before opening the release PR if no matching section exists**, so this has to be on `main` before dispatch.

Proposed bump: **minor** (2.12.0 → 2.13.0). Three items change what a gate reports, which is more than a patch:

- a security scanner that fails to start now **WARNs instead of FAILs** — a repo missing bandit/semgrep/pip-audit will stop failing that gate
- findings across byte-identical file copies are **collapsed**, so reported counts drop on repos that vendor or distribute duplicated trees
- `sm doctor` version drift **warns instead of blocking** a refit

The section calls out explicitly that **hull letter grades are unchanged**, since the Action's `minimum-grade` input depends on them — only a findings count and delta were added to the label.

Everything except the `mcp` advisory came from the rulebook-ai run; ten items were filed as barnacles against this repo.

Once this merges I'll dispatch **Actions → Release → minor**.

## Summary

Adds the `2.13.0` release section to `CHANGELOG.md`.

- Documents scanner, duplicate finding, doctor, hull grading, and refit changes.
- Includes fixes and advisories, including the `mcp` advisory.
- Enables the release workflow to create the release PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,10 +53,13 @@ against this repo. The case-study artifacts are in `DOCS/case-studies/`.
   project path. Paths are now project-relative, with both spellings remapped
   and symlinked roots resolved on both sides.
 - **`dangling-references` could not tell code from prose** (#332) — Python
-  subscript-then-call is byte-identical to markdown link syntax, so a
-  documented `handlers[key](arg)` inside a fenced block was reported as a
-  broken link. Fenced blocks are now skipped and inline code spans blanked,
-  with multi-backtick and line-wrapping spans handled.
+  subscript-then-call — indexing a dict of handlers, then calling the result
+  — is byte-identical to markdown link syntax, so a code sample using that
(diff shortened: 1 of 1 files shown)


ScienceIsNeato:
[fixed_in_code] Fixed. You're right that byte-identical was wrong — they share delimiter structure, not bytes. Rather than soften it to 'resembles', the entry now says it matches the inline-link pattern exactly, which is both accurate and the reason the gate misfired. Also corrected the same phrasing on the case-study page.

ScienceIsNeato:
[fixed_in_code] Fixed. 'The latest release' stops meaning anything the moment a newer semgrep ships, so the entry now names semgrep 1.172.0 and states the trigger outright: the ignore must be re-checked when semgrep unpins mcp.

